PR #54, opened 2015-08-05
Improved Partition ID Handling

miszr:
I cleaned up and fixed code for handling partition IDs which now work properly with RAID.

The issue was caused because the name of RAID devices in Linux is in the /dev/md{%d}p{%d} format which caused the image files to be incorrectly named on the server(eg d1pp1).

### DIFF
--- a/src/buildroot/package/fog/scripts/bin/fog.download
+++ b/src/buildroot/package/fog/scripts/bin/fog.download
@@ -184,7 +184,6 @@ elif [ "$imgType" == "mpa" ]; then
 fi
 if [ "$imgType" == "n" ]; then
     parts=`fogpartinfo --list-parts $hd 2>/dev/null`;
-    diskLength=`expr length $hd`;
     if [ -f "$win7imgroot/sys.img.000" ] && [[ "$osid" == +([5-7]|9) ]]; then
         if [ "$win7partcnt" == "1" ]; then
             restorePartition "$win7sys";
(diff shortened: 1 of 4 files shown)


mastacontrola:
Quick question.

All of these edits, and normal imaging still works?

miszr:
I have tested all types of imaging(Upload, Download, Multicast) on a system with RAID(The reason for the changes is because of multicasting did not work because of the incorrectly named image files.).

The old code did the following:
```bash
diskLength=`expr length $hd`;
partNum=${part:$diskLength};
```
So if `$hd` is `/dev/sda` and `$part` is `/dev/sda1` the code removed everything related to the actual disk.

The problem that occurred is when you had a RAID drive:
`$hd` is `/dev/md126` and `$part` is `/dev/md126p1` which resulted in `partNum="p1"`, which caused the image name on the server to contain an additional 'p' in the file name which caused the scanf code in `MulticastTask.class.php` to be unable to parse the file name.

The new code just takes the last number in the `$part` string and uses it as the partition number which means the code no longer relies on `diskLength` at all.

fractal13:
So, the only effect of these changes is to calculate the partition number as the last digits in the partition device name.  Correct?

There are a few cases that aren't caught, because they aren't using the "diskLength" method.

miszr:
@fractal13 Yes it only takes the last digits.

Thank you for pointing that out. I have added an additional commit to hopefully handle all those cases.

fractal13:
How does the the RAID array get rebuilt?  Is the information stored in the MBR area of the disk, or do you have to manually build it before using FOG to download an image?

miszr:
@fractal13 Software RAIDs(Intel Raid included) store information about the raid in the superblocks of the RAID disks.

The problem only occurs for Software RAIDs, since hardware RAIDs show up as a normal disk devices in the operating system.

If you look at the "RAID Support" pull request i submitted this week you will see that the `mdadm` command automatically detects and assembles the RAID device.

miszr:
@mastacontrola @fractal13 Any more questions?

mastacontrola:
Just adding so you know we haven't forgotten about this.

fractal13 is working on another approach that does basically the same thing in a much more dynamic/generalized state.  This is the ONLY reason I have not merged this item yet.

mastacontrola:
Closing this pull request as this is sorely outdated and we have moved to a much more improved partition handling scheme.

PR #55, opened 2015-08-10
Fixed mismatching types in comparison during CentOS install

miszr:
Old comparison caused "integer expression expected" error.


mastacontrola:
Adding Thanks for the help.

PR #67, opened 2015-10-20
backup the configs even if one some of the folders is read only

vipconsult:
I found the problem

I am running fog in a docker container and the mv command was failing
the folder /var/www/fog/service/ipxe/custom is mounted as a docker volume from the host so the content is read/write , but the folder (custom ) cannot be deleted so the "mv" command was failing.

now it does cp and rm instead. I hope you except the change.


mastacontrola:
Thanks for submitting.  I've made a few slight adjustments that will be synced here, but ultimately it's the same thing so thank you for your contribution.

PR #76, opened 2016-01-06
Accurately calculate the start range for DHCP.

wayneworkman:
Wrote a function that accurately adds a number to an IP - intended to add a base number to the network address. I left it at 10 - but now it's accurate.


wayneworkman:
IFS=$PreviousIFS  needs a lower case "p"  sorry about that.

PR #87, opened 2016-02-26
Accurately suggest best IP and Interface, recalculate subnet mask if …

wayneworkman:
…alternative is given, plus display the subnet mask before proceeding with installation.


wayneworkman:
if an alternative interface is given, subnet mask is recalculated. Still need to improve the suggested interface - because if the IP doesn't match what was previously stored as suggested, then the suggested interface needs recalculated based on the given IP.

I can get this done tonight - but for now,  this code is pretty solid and good to go.

PR #91, opened 2016-03-10
Commited Sebastian Roth's suggested fix for purposly making DHCP fail…

wayneworkman:
… if an invalid ROUTER or DNS address is provided from the .fogsettings file.


mastacontrola:
I'm closing this as while your issue was "valid" fixing by making the program error out due to pure syntax is kind of the wrong approach.  The problem was not that the fields you expected to be there "MUST" be there because they simply do NOT have to be there.  In your case it broke because your environment does need these values and the .fogsettings file did not contain the appropriate IP addresses as they also contained text.

I've pushed what I hope will fix this for you by simply checking the variables if they're valid ip's or not.  If they're not it will attempt to obtain the ip from the variable and then try again.

mastacontrola:
Added bug/enhancement to the labels as this was indeed a bug and (I think) an enhancement of how it should operate.  I added the invalid because the request is trying to say something MUST be there when that is not the case at all.

PR #93, opened 2016-03-20
Moved next-server option from global config to subnet config.

wayneworkman:


wayneworkman:
This makes sense to me, and it's how I've been running DHCP here at the house for about half a year. I have a multi-nic server that serves as a router, it's connected to two different networks. I don't want my next-server leaking onto the outside network due to it being a global config. 

Also, moving it into the subnet greatly simplifies the fogDHCP plugin I'm working on. In this plugin, I only want to enable users to easily manipulate subnet listings, not global configs. Moving the next-server into the subnet listing simplifies this.

PR #102, opened 2016-06-01
Add changes for postinstall scripts

ghooper90:
Added two new pages one is for exposing fog database values to the post install scripts to allow more robust post install actions. The second is to allow (re)setting of the target host name from within the fog post install script. The hostinfo.php file requires just a mac address. The sethostname.php file requires the mac address, old host name and new host name to complete successfully.

### DIFF
--- a/packages/web/lib/fog/hostinfo.class.php
+++ b/packages/web/lib/fog/hostinfo.class.php
@@ -1,0 +1,55 @@
+<?php
+class HostInfo extends FOGBase {
+    protected $macSimple;
+    protected $repFields = array(
+        'hostName' => 'hostname',
+        'hostDesc' => 'hostdesc',
+        'imageOSID' => 'imageosid',
+        'imagePath' => 'imagepath',
+        'hostUseAD' => 'hostusead',
+        'hostADDomain' => 'hostaddomain',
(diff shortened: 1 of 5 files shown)


mastacontrola:
There's only one part I'm not very fond of.  This is the "setting the hostname".

While I am more than willing to add the exposure code (so your post download scripts will work), I'm not going to add the setting the hostname as this can expose too much API to untrusted/unknown users and it doesn't require validation.  While you've placed some good snippets to kind of help prevent it from "just occurring" I just feel it's too much a risk to the DB at that point.

I'm planning to make some minor adjustments as well.  Remember, as you're already collecting the host information, why not just use the existing code base in it?  We already have the hosts get loaded with their relevant information on URL calls anyway.  No need to call subsequent queries to get the data.

mastacontrola:
@ghooper90 Thank you for the code snippets.  While I didn't merge the code, I did use your code base and implement what you needed, with a bit more dynamically usable content.  I'm not adding the setHostname part as I already informed you I'm not a fan of exposing api to make changes without validation of who is logged in to make the change.

Hopefully you'll find this is suitable though.

Thank you for the time on this though.  It greatly helped me out to get something for you.

PR #109, opened 2016-07-02
Add automatic correction of storage node credentials.

wayneworkman:
This change applies to both the main server and storage nodes.
It queries the DB for the current node's credentials, and compares to what the installer has set in the OS. If they do not match, a banner is displayed informing the user that they will be corrected. Then, they are corrected.

### DIFF
--- a/bin/installfog.sh
+++ b/bin/installfog.sh
@@ -522,6 +522,7 @@ while [[ -z $blGo ]]; do
                         echo
                     else
                         registerStorageNode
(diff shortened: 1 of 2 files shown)


Nekroze:
This would be quite handy as I have had issues importing a database backup via the web interface and suddenly upload will not work. Took me ages to figure out that the import changed the master nodes FTP password. Forgive me if this is not what you are talking about.

PR #114, opened 2016-07-27
Added Surface Pro 4 and Apple Intel Mac DHCP configuration.

wayneworkman:
The Apple stuff uses the stuff that Sebastian made and had in the wiki for "FOG on a MAC" in the "Fancy" section. I've not tried it on a Mac myself, but I know Sebastian has. Also, I did run this configuration on my DHCP server and the dhcpd service was able to restart without issue, so it won't break dhcp.

The Surface pro 4 part uses the only known vendor class identifier - and is likely sub-model specific. I am almost 100% sure it will work correctly for this very specific model, but don't have a surface to play with. I did put the configuration in my DHCP server and it was able to restart without error. One class is better than none, and it will only match to that exact model, and will give pass the name of the only working efi file we have for Surface Pros. We need to try very hard to collect as many vendor IDs for Surface Pros of all models and sub models as possible so that we can have a DHCP class for each, to be able to give the best support for them out-of-the-box for FOG 1.3.0.

### DIFF
--- a/lib/common/functions.sh
+++ b/lib/common/functions.sh
@@ -1929,6 +1929,28 @@ configureDHCP() {
             echo "        match if substring(option vendor-class-identifier, 0, 20) = \"PXEClient:Arch:00009\";" >> "$dhcptouse"
             echo "        filename \"ipxe.efi\";" >> "$dhcptouse"
             echo "    }" >> "$dhcptouse"
+            echo "    class \"SURFACE-PRO-4\" {" >> "$dhcptouse"
+            echo "        match if substring(option vendor-class-identifier, 0, 32) = \"PXEClient:Arch:00007:UNDI:003016\";" >> "$dhcptouse"
+            echo "        filename \"ipxe7156.efi\";" >> "$dhcptouse"
+            echo "    }" >> "$dhcptouse"
+            echo "    class \"Apple-Intel-Netboot\" {" >> "$dhcptouse"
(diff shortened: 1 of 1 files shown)


Quazz:
On that note, if we're doing this anyway, couldn't we add the BIOS/UEFI compatible DHCP stuff as well?

mastacontrola:
Why not, in the meantime, just switch all efi files to use the 7156 labeled form?

We know it works for all things up to this point at least.  Why would we start designating different efi based files based on model.  Use what we know works (in the code base) and down the line ask users to test with the non-version labelled files.

This way we're not having to play a big ol' guessing game?

wayneworkman:
@mastacontrola Sounds good - but I just hope that it will work with the very latest models without issue. Perhaps we name the old file to just `ipxe.efi` and then have a newipxe.efi file. Then, let that ride for a while and see if it works for everyone. Also - maybe we should really try to figure out what's different between the two files?

@Quazz That stuff is already in there, has been for a long while - it was just lacking the Apple stuff.

Quazz:
Ah, I'd been using manual for so long I didn't even realize, my bad.

mastacontrola:
I've added the code but only to the working-RC-5 group.

Normally, yes, you would make a pull request under dev-branch, but seeing as we're now in RC status and dev-branch is where things default to, the dev-branch is, more or less, on lockdown.

I'm closing the pull request without merging it but that's only because the relevant code is posted under the working-RC-5 branch.

Hopefully you all understand.

PR #132, opened 2016-09-28
Fixed ipaddress variable listing @line 336

RichardRay:
When running the installer on a machine with multiple 'good' interfaces the function getFirstGoodInterface returns all interfaces, line 336 takes the output of getFirstGoodInterface and uses the awk command to define this variable ($ipaddress) with the resulting string.

which returns a string of ALL IP ADDRESSES delimited by spaces.

further down this variable is used to write to the httpd/conf.d/fog.conf which breaks the installer entirely

my idea is just to cut the resulting variable to just use the first entry in that string, this could also be accomplished by rewriting the getFirstGoodInterface function to actually... get the FIRST good interface.

### DIFF
--- a/bin/installfog.sh
+++ b/bin/installfog.sh
@@ -333,7 +333,7 @@ exitcode=$?
 [[ -z $bldhcp ]] && bldhcp=""
 [[ -z $installtype ]] && installtype=""
 [[ -z $interface ]] && interface=$(getFirstGoodInterface)
-[[ -z $ipaddress  ]] && ipaddress=$(/sbin/ip addr show $interface | awk -F'[ /]+' '/global/ {print $3}')
+[[ -z $ipaddress  ]] && ipaddress=$(/sbin/ip addr show $interface | awk -F'[ /]+' '/global/ {print $3}') && ipaddress=$(echo $ipaddress | cut -f 1 -d " ")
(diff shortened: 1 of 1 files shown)


mastacontrola:
Just trying to understand, but is the file in the bin folder called goodInterface.txt containing all the interfaces or is it only showing a single interface?

mastacontrola:
I'm thinking more along the line 319 in lib/common/functions.sh to be changes like:
`goodInterface=$(cat $workingdir/goodInterface.txt | head -1)`

mastacontrola:
Something more along the lines of: https://github.com/FOGProject/fogproject/commit/bd6855c6b26ff39aadbe7310ec1e640662158dc4

RichardRay:
to answer your first question I think goodinterface.txt is deleted in the getFirstGoodInterface function. I've also removed the install log files since I was able to install successfully.

and yes that is a damn fine method of dealing with it, I didn't want to alter anything in functions.sh unless I had to and this is just the solution that worked for my install. 

That looks good to me :) So long as this issue doesn't affect anyone else I'm beyond happy.

mastacontrola:
Then I'm going to close the pull request.  Thanks for the info and help.

mastacontrola:
And you're correct it does remove the file, I was thinking more along the lines of breaking before it deletes first then check.  But I think this will work as is now.

PR #138, opened 2016-10-17
Install php 5.6 by default if dnf isn't present.

wayneworkman:
This fixes an issue came in at commit number: [af73066572bfbb9108b31fcb01c4b4ac821c4429](https://github.com/FOGProject/fogproject/commit/af73066572bfbb9108b31fcb01c4b4ac821c4429)

That commit fixed the Fedora 22+ installation (uses dnf), but it also altered the installation unnecessarily for other distributions (doesn't use dnf). This fix undoes the changes to the other distributions.

### DIFF
--- a/lib/redhat/config.sh
+++ b/lib/redhat/config.sh
@@ -17,7 +17,7 @@
 #   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 command -v dnf >/dev/null 2>&1
-[[ $? -eq 0 ]] && repos="remi" || repos="remi,epel"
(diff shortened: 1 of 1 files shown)


mastacontrola:
I found the issue and have it more properly addressed.

The issue is specifically caused because of the installation of the repos as needed.  the installers were requiring the exact same repositories that it's trying to install.  Because these don't exist at that point the installer was failing to install the epel and remi repos.

Found the issue and addressed it more properly.  Thanks for taking the time to edit and find a portion of the problem.

I've more or less added the fix here among others.

PR #143, opened 2016-11-25
Fix hanging upload task

matteodelabre:
We've recently had a problem while capturing images from newer machines. The upload process would hang at the following message :

```sh
* Saving original partition table...
```

I managed to track down the problem to the `getPartitionTableType` function in `partition-funcs.sh`, and more specifically to the following line:

```sh
local mbr=$(gdisk -l $disk | awk '/^\ *MBR:/{print $2}')
```

So I booted with a debug task and ran this command in interactive mode. On normal machines, the program answered as expected. However, on the newer machines, probably due to the fact that both GPT and MBR are present, the command asked a question that the `fog.upload` script never answered.

I fixed the problem by piping `yes ''` into the command so that any question gets answered with the default answer.

### DIFF
--- a/src/buildroot/package/fog/scripts/usr/share/fog/lib/partition-funcs.sh
+++ b/src/buildroot/package/fog/scripts/usr/share/fog/lib/partition-funcs.sh
@@ -538,8 +538,8 @@ processSfdisk() {
 getPartitionTableType() {
     local disk="$1"
     [[ -z $disk ]] && handleError "No disk passed (${FUNCNAME[0]})\n   Args Passed: $*"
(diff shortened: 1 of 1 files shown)


mastacontrola:
Much appreciated, but I think the problem typically stems from Windows OS reinstall that converts GPT to MBR, without removing all the proper bits.

We do have a function that should be able to correct this (more appropriately) using `fixparts`.

While I'm fairly good, I haven't figured out a nicer mechanism to break out of the input and perform the fixparts as expected, so maybe this will prevent people from having the problem altogether?

mastacontrola:
Updated init's on the main server, so a re-install will have these corrected changes in place.

I will probably need to look at them a bit further on to fix this in a better fashion, but thank you so much for the time.

PR #144, opened 2016-11-29
Fix detection of systemd under Ubuntu

LukeCarrier:
It looks as though the installer expects a value of `"yes"` in the `$systemctl` variable if systemd is configured as the init system. The current implementation in `/lib/ubuntu/config.sh` results in `$systemctl` being assigned the path to `systemctl` (e.g. `/bin/systemctl`) instead, causing later comparisons to fail.

This may have gone unnoticed for some time as it appears both tools can be installed simultaneously. Newer Ubuntu releases don't seem to ship with `sysv-rc-conf` installed, causing enabling the Apache/PHP-FPM services to fail, thus preventing installation.

### DIFF
--- a/lib/ubuntu/config.sh
+++ b/lib/ubuntu/config.sh
@@ -34,7 +34,7 @@ if [[ $linuxReleaseName == +(*[Bb][Uu][Nn][Tt][Uu]*) ]]; then
                 rm -rf /etc/php* /etc/apache2*
                 echo "Done"
                 dots "Stopping web services"
-                systemctl=$(command -v systemctl)
+                command -v systemctl && systemctl="yes"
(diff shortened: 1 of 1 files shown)


mastacontrola:
Hi @LukeCarrier Thank you for noticing and attempting to fix.

I was made aware of these issues over the weekend and have addressed the systemctl a bit more properly:

https://github.com/FOGProject/fogproject/commit/c31bebe226668f860bcf6812fc5995c35a0f2d1f

This is what's going to be released for the RC-27 public push.  Your approach would also work, but I think having a more generalized method and approach would be more appropriate.  By moving all of the dealings into a single file we can have less impact on when OS's switch different things around on us as well.  Hopefully you agree and understand that I truly appreciate the effort and understand why I'm not merging this particular "fix".

LukeCarrier:
@mastacontrola understood -- your approach makes a lot more sense.

Thank for your the prompt response to the PR and your work in general, it's very much appreciated. :)

PR #183, opened 2017-09-09
Fix installfog.sh force-https option

Sebastian-Roth:


Sebastian-Roth:
Tested on CentOS7 with already installed FOG and existing .fogsettings.

Sebastian-Roth:
Also tested in a fresh clean Ubuntu 16.04.3 server system using `./installfog.sh --force-https`.

PR #185, opened 2017-09-10
Make initrd handle SSL properly

Sebastian-Roth:


Sebastian-Roth:
Tested full and quick register, quick delete, deploy. Should be all fine I hope. Please take a look Tom and see if you like the change. Great if you could build fresh inits from that as well.

mastacontrola:
Building init's right now. will update when complete.

mastacontrola:
Init's built and updated. RC-9 forthcoming.

PR #191, opened 2017-09-24
Added system UUID to the POST parameters to register fully

Sebastian-Roth:


mastacontrola:
Whoops thanks for noticing and fixing.

PR #196, opened 2017-10-20
Disable sysuuid detection code for now as it is causing trouble

Sebastian-Roth:
with machines having the same UUID like we've seen on some MSI
motherboards all showing FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF

### DIFF
--- a/packages/web/lib/fog/fogbase.class.php
+++ b/packages/web/lib/fog/fogbase.class.php
@@ -516,17 +516,21 @@ abstract class FOGBase
                 $mac = filter_input(INPUT_GET, 'mac');
             }
         }
-        $sysuuid = filter_input(INPUT_POST, 'sysuuid');
+// disabling sysuuid detection code for now as it is causing
+// trouble with machines having the same UUID like we've seen
+// on some MSI motherboards having FFFFFFFF-FFFF-FFFF-FFFF...
+/*        $sysuuid = filter_input(INPUT_POST, 'sysuuid');
         if (!$sysuuid) {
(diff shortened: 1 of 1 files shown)


Sebastian-Roth:
Sorry Tom but I will just merge this right away as I'd like others to test this as well. Have been working on this the whole evening to figure it out.

mastacontrola:
No need to be sorry you’re a developer too!

pcnate:
How would one go about checking the UUID on a system. We have a client that has many very cheap ByteSpeed machines and I know for a fact that they don't set many values so I don't expect the UUID to be set in these.
Another question is would the iPXE loader have the ability to set a UUID when this happens?

Sebastian-Roth:
@pcnate There are many tools available in [windows](http://www.waynehoggett.com/2009/08/get-local-computer-uuidguid-using.html) and [linux](https://stackoverflow.com/questions/10152762/best-way-to-get-machine-id-on-linux) to get the UUID. All those **should** match the UUID that iPXE reads from the SMBIOS.

No I don't think it's possible for iPXE to set UUID for you. iPXE would have to modify your BIOS firmware. Though there are tools around to do that. But probably not for cheap hardware I suppose. Read through [issue 198](https://github.com/FOGProject/fogproject/issues/198) as well and pay attention to the latest comment I just added.

We are working on a solution (on a [separate branch](https://github.com/FOGProject/fogproject/tree/unique-identifier)) to this by adding more unique parameters to identify clients reliably. But it's been a struggle as we are all very busy at the moment.

PR #205, opened 2017-10-27
Add bitlocker detection before image upload

Sebastian-Roth:


Sebastian-Roth:
This seems fairly simple from what I have found out about this. It would have been more complex if we'd want to detect Bitlocker To Go as well but AFAIK that is only done on removable media which we don't care about.

See what you think of it (untested so far!) and please build new inits with it for people to test. Thanks!

mastacontrola:
Files are updated and uploaded.

PR #209, opened 2017-11-01
 Implement a first good sketch of the unique host identification

Sebastian-Roth:


mastacontrola:
I only looked over the PHP based stuff. You may want to get @FinancedHydra to take a look at the client based code for review.  This way we can merge this commit.  I really love where this is headed to be honest though.

Sebastian-Roth:
@mastacontrola Thanks for the comment and your thoughts on this. I have done some first simple tests but I have no idea if this will work in all cases. Though this seems fairly straight forward. I think we need to get people to test this. So I will add this to `unique-identifier` branch of the official FOG project now and ask people to go on this branch an test.

Issue #195, opened 2017-10-18
Detect Windows boot/recovery partition


Sebastian-Roth:
So far we use the [filesystem labels](https://github.com/FOGProject/fogproject/blob/dev-branch/src/buildroot/package/fog/scripts/usr/share/fog/lib/funcs.sh#L458) (**remember to fix the wording in the script code**) to find boot/recovery/reserved partitions on Windows systems. This fails on systems having a different label, namely foreign language installations. See [here](https://forums.fogproject.org/topic/10963/win7x64-golden-image-partitions-problems), [here](https://forums.fogproject.org/topic/10306/windows-10-resizable-image-system-partition-issue) and [here](https://forums.fogproject.org/topic/10485/best-way-to-make-windows-10-master-and-upload).

We need a different way of detecting those "special" partitions.

mastacontrola:
I've added a label for the French "Reserved" label, however I agree that a much better method of detecting reserved partitions should be created.  There's too many different languages for us to just keep on "adding" new labels to search for. Also, labels can be set by the user, so it's kind of an unreliable mechanism in some situations.

Sebastian-Roth:
Trying to gather more information on this and discussing this with others we came up with four different types of partitions so far:
1. Windows **system partition**\* - This partition contains the files need to boot up Windows and should be **fixed in size and start sector**. Should be fairly easy to detect.
   1. Windows Vista/7/8/10/Server 2008: `Boot\BCD` (registry hive config file) and `bootmgr` (boot loader binary)
   2. Windows XP: `boot.ini` (INI config file) and `NTLDR` (boot loader binary)
2. Windows **recovery**, EFI and OEM partitions as well as Linux SWAP and `/boot` partitions - We want those to be **fixed in size but have a movable start sector** as those partitions are last on disk quite often. So we need to move those to be able to expand the Windows partition.
   1. Windows **recovery** - should be easy, find `Recovery\WindowsRE\winre.wim` file.
   2. EFI partition - easy, find `EFI\Boot\BOOTX64.EFI` bzw. `EFI\Boot\BOOTX86.EFI`
   3. OEM partition - Almost impossible to detect as every vendor does his own thing.
   4. SWAP is easy to detect using `blkid`
   5. Linux `/boot` shouldn't be to hard as almost all distros use GRUB these days.
3. Windows **boot partition**\* - On systems with only one single partition for boot and system we want this single partition to be **fixed in start sector but not fixed in size**.
    1. TBD
[...]

PR #210, opened 2017-11-07
Fix AWK foo to not move fixed size partition starting sectors

Sebastian-Roth:


Sebastian-Roth:
@mastacontrola Please take your time and have a close look at this before merging it into the working branch. The issue was introduced with [this commit](https://github.com/FOGProject/fogproject/commit/ef75aef1547f17686c65a2d3a7653e111507ac44#diff-def42937359ff06f0443fffc3a53df12). Do you remember why the logic was changed that way? I am not saying my new proposal is all good. I tested with the issue case we have in that [forum post](https://forums.fogproject.org/topic/11041/deployed-image-boot-loop-black-screen-press-ctrl-alt-delete) and it's fine from my point of view. But I am not sure if this could cause another different issue now.

Sebastian-Roth:
Is there a good reason for even having such a crude fixed list like `1:3:5` I am wondering? How I understand fixed is that those partitions should not be moved at all because it can cause an issue. So what's the point of having 2:4 as resizable but not being able to expand those because 5 (and 3) cannot be moved?!

mastacontrola:
The reasoning being is we've seen some people push their recovery partition to the end of the disk.  The main data partition would only be able to expand until the "fixed start" in this scenario.

Sebastian-Roth:
Well I think our current implementation is kind of to minimalistic. We use the term "fixed" for two different things. One is "fixed size" - probably where it originated from. Though I am not sure if we actually need that I am open for a discussion on this. Why do we need partitions to be fixed in size but are still fully working even if they are moved around. Is that because they have filesystems that we cannot resize?

And the second "fixed" that we have mixed into this is "fixed starting position" which is important for boot partitions (being referenced by a static pointer).

I'd love to entangle those two things and have a new `d1.fixed_start_partitions` list added. Though I have to admit that I haven't looked into the code yet. So I cannot say how big this change would be. What do you think @mastacontrola?

fractal13:
The fixed size partition is to support things like linux swap partitions that probably shouldn't scale with the size of the drive, but are reflective of RAM.  What about Microsoft's and Apple's recovery partitions.  Don't those just have a fixed size need due to the software tools that are stored on them?  I wouldn't want to grow them just because I go to a larger disk.

I believe that was the original reasoning in my mind.

Sebastian-Roth:
Thanks heaps for your reply @fractal13. This kind of supports my theory that fixed size was originally meant to be just that fixed in size. We use it as fixed in start position and fixed in size now. But sometimes we don't really stick to the fixed starting point and therefore break things.

I'll try to come up with a proposal to separate those two things and have a very own list for each: `dX.fixed_size_partitions` and `d1.fixed_start_partitions` (the later being fixed in start and most probably in size as well).

Sebastian-Roth:
One major first step to achieve what I am aiming for would be to properly detect those different types of partitions. So as a preparation for this I am trying to collect the information here: https://github.com/FOGProject/fogproject/issues/195

mastacontrola:
@fractal13 You're correct in the thoughts. Recovery/Reserved partitions should be fixed in size, but where things are starting to head in here is that the fixed size partition assumes the start position of the fixed partition is also unmovable. The unmovable start point should only be on the "boot" volume, whatever that may be.

The reason I'm so cautious on this is some users have their windows layouts set so the Main data partition occurs BEFORE the recovery partition. The moving of start position would be required for the recovery partition, while the total size of the recovery part would remain the same as it was when we captured it.  That's the idea, and I think our current AWK script is VERY close to making this happen appropriately. The only issue, as I'm seeing it currently, is consecutive fixed partitions were the adjacent partition start point is shifted although the prior was fixed, the start position should be identical, or so I thought.

PR #216, opened 2017-11-24
Fix issue with restore swap uuid on mbr based hdd's

m1chalm:


Sebastian-Roth:
@m1chalm Thanks heaps for looking into this an sending a pull request! From my point of view your change might bring in a new issue on GPT disks. This initial logic was to only set `parttype=82` if the SWAP UUID is present (non zero string).

Please change your pull request to match that logic and we shall add it soon.

m1chalm:
@Sebastian-Roth Sorry about that. I thought that this is another issue. This logic causes that if swap uuid is not present (zero string variable uuid), swap partition will not be restored (created).

But, if you say that this is correct, I will push changed code in few minutes.

Sebastian-Roth:
@m1chalm Not a problem at all. Thanks for clarifying why you wanted to change that too. This part of the code logic has been there for a long time. Looking through the code history I found this comment:
```
        # don't have a good way to test, as ubuntu installer
        # doesn't set the GPT partition type correctly.
        # so, only format as swap if uuid exists.
```
So possibly this is something from the old times and could be removed but on the other hand it has not caused any issue in the last three years and I'd say we should leave that check in place.

Thanks again for your quick response. I will merge this into our working branch now.

PR #225, opened 2018-01-18
Change color of deploy/upload to avoid error

Shiigehiro:
Add different color for upload and download (similar) icons

### DIFF
--- a/packages/web/management/css/default/fog.css
+++ b/packages/web/management/css/default/fog.css
@@ -244,6 +244,13 @@ input[type=checkbox].delvid:checked + label:before {
     display: block;
     height: 18px;
 }
+
+#search-content>tbody>tr>td>a>.icon.fa.fa-download {
+  color: #7baa0f
+}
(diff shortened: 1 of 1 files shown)


Sebastian-Roth:
@Shiigehiro Thanks for your pull request. I see what you are aiming for and I think it's quite a nice idea. But your code change doesn't work for me. Maybe like this:
```
.icon.fa.fa-download {
    color: #7baa0f
}
.icon.fa.fa-upload {
    color: #ff9c07
}
```
Note the indent style as well.

Shiigehiro:
@Sebastian-Roth This work for me, so yes, this work too with this, just more global, i doesn't know if those icons are used for another thing in FOG

Shiigehiro:
@Sebastian-Roth I will remake all the french translation too, because more than a half is false

Shiigehiro:
Can you pull it ? :o

mastacontrola:
Thank you much for the contribution.

PR #227, opened 2018-01-26
Safely handle all special characters in MySQL password

blkeller:
In the current dev-branch, if the user supplies a MySQL password to the installer that contains special characters, the installation fails in various ways.  Depending on the characters present, sometimes failure is obvious immediately upon database schema installation, and sometimes it is less obvious and only seen as a DB lockout after the first FOG upgrade installation due to generational corruption of the stored password.

This pull request adds proper escaping of special characters within the user-provided MySQL password as needed for the four different escaping schemes used by bash, sed, PHP, and MySQL itself.

To test these patches, I used a worst-case-scenario MySQL password of:
`{}It's; ""a" \whacky*/ (''MySQL) \\P@ss\'word![&]`

This password causes the current dev-branch FOG to choke on install.

With these patches, I was able to complete installation and access the FOG server's web GUI without issue.  I then performed two re-installations on top of this existing test installation to simulate two subsequent FOG upgrades.  This was to test for generational password corruption issues.  After both re-installs, I was still able to access the web GUI as expected.  I also verified that the `snmysqlpass` line in `.fogsettings` and the `DATABASE_PASSWORD` defined in `config.class.php` had been generated by the installer with proper escaping and didn't change between re-installations.

### DIFF
--- a/lib/common/functions.sh
+++ b/lib/common/functions.sh
@@ -1592,34 +1592,34 @@ configureHttpd() {
         dummy=""
         while [[ -z $dummy ]]; do
             echo -n " * Is the MySQL password blank? (Y/n) "
-            read dummy
+            read -r dummy
             case $dummy in
                 [Yy]|[Yy][Ee][Ss]|"")
                     dummy='Y'
                     ;;
                 [Nn]|[Nn][Oo])
                     echo -n " * Enter the MySQL password: "
-                    read -s PASSWORD1
+                    read -rs PASSWORD1
                     echo
                     echo -n " * Re-enter the MySQL password: "
-                    read -s PASSWORD2
+                    read -rs PASSWORD2
                     echo
-                    if [[ ! -z $PASSWORD1 && $PASSWORD2 == $PASSWORD1 ]]; then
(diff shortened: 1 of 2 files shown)


mastacontrola:
Thank you very much @blkeller for taking the time and thoroughly testing your changes. I'm going to merge them into the dev-branch as is, merging dev-branch back into working.
For future reference, please try to make pull requests to the working branch as this is where we are making new changes quite often. Either way, Awesome and thank you so very much.

blkeller:
Thanks for the merge, @mastacontrola!  My organization has been using FOG for several years for imaging, so I'm happy that I was able to give back to the project even in some small way.  Looking forward to rebuilding our old but reliable 1.2.0 server and upgrading to 1.5.0 so that we can start using the FOG client for configuration management and inventory management.

Re: working branch vs. dev-branch branch, sorry if that caused any issues for you.  I'll make a note to myself to use the working branch for any future PRs.  I had gone with dev-branch based on the project's `README.md`:

> If you would like to create a pull request, please make the pull request into the dev-branch branch.

You may want to consider updating that text for others' reference if the working branch is more appropriate.

Thanks for all you do for FOG!

PR #239, opened 2018-03-16
Changes to ipxe to allow for arm* architectures

parkerberb-nfv:
This assumes that refind_aa64.efi and grub_aa64.exe exist but does not add or build them. This is just a small configuration change to allow the arch to be set as arm* and then change the bootloaders accordingly.


parkerberb-nfv:
@mastacontrola @FinancedHydra 
Please review and let me know what you think! :)

PR #260, opened 2018-08-15
Respect FOG_UDPCAST_INTERFACE

hanazuki:
Use FOG_UDPCAST_INTERFACE setting for outgoing multicast packets, instead of dynamically finding interface (which does not work well).

### DIFF
--- a/packages/web/lib/service/multicasttask.class.php
+++ b/packages/web/lib/service/multicasttask.class.php
@@ -42,11 +42,7 @@ class MulticastTask extends FOGService
         if (!$StorageNode->get('isMaster')) {
             return;
         }
-        $Interface = self::getMasterInterface(
-            self::resolveHostname(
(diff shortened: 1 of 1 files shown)


hanazuki:
Sorry, I've created PR against different repo by mistake.

mastacontrola:
I understand what you're doing here, but I don't understand the why.  The functionality you're changing is intended to ensure the interface is not just a user value (which can be typo'd or incorrect for the machine you're working off of.)  Essentially the part you're changing in the code is designed to take the node's ip address and find the interface dynamically. So this way, let's just say, you change the node's NIC to eth14 when it was eth0.  The way it works currently is it will find the interface for the IP associated to the node. So an update would only require the IP address to be changed, not the node's interface (which is only used for the bandwidth page) and certainly not the global setting (which doesn't really make much sense in a multi-node setup.)

PR #267, opened 2018-10-15
Remove sysv-rc-conf as it is not part of Ubuntu 18.04 anymore

Sebastian-Roth:
and causing problems on upgrade if not fully purged


Sebastian-Roth:
Reference: https://forums.fogproject.org/topic/12528/sysv-rc-conf-missing

PR #272, opened 2018-11-24
Improve web UI interaction by decoupling some web requests

Sebastian-Roth:
Hi @mastacontrola, you know I have been looking into some web UI performance issues. I have changed a fair bit of code in various places. See what you think of it. Please comment if you feel like it.

* Main issue at first seemed to be the master node loading names of logs/images/snapins from the storage nodes even if the information is not needed on that page. I know generalizing code is great but from my point of view this went too far and so I added a new API route to load those details of needed. You might want to update the [news post](https://news.fogproject.org/simplified-api-documentation/) if we decide to keep it like this.
* As I had changed this I noticed that other calls took very long as well (I simulated lower bandwidth links and even miss-configurations to the what's happening). Digging much deeper I found that many of those calls to http://.../fog/status/...php open up a database connection without actually using it. Single ones wouldn't hurt but there are many of them used when browser the web UI on the master and by fog-clients as well. I guess I still haven't found all of them but things seem to have improved immensely already. :-)
* I guess function `establish()` in lib/db/databasemanager.class.php could take even more work on. The schema redirect kicks in if for example the DB connection from storage to master cannot be established (for whatever reason). Then triggers another request (and DB load) which is definitely not what we want.
* Fixed the interface enumeration for bandwidth monitor - just a typo.
* I removed the call to get the external IP (http://ipinfo.io/ip) as I don't think this is of any use in pretty much any environment I can think of. Here is a [reference](https://github.com/FOGProject/fogproject/commit/d3c2a9616ab71daac4c3b6a70bcb05b4fd57e333#diff-ddd422e457584565149f1262d78d4381) to the commit that added the code. From the comment it seems like something about the services. Still don't think we need it.

That's about it. See what you think.

Edit: I have done a fair bit of testing, web UI ok, boot menu ok, capture ok, deploy ok, multicast ok, fog-client and other things **untested**

Edit: I pushed the multicast changes (67b4e64) into this branch as well as I feel this is adding to the speed improvements as well. At least in huge environments with a dozen or so storage nodes this definitely makes a difference.

### DIFF
--- a/packages/web/lib/db/databasemanager.class.php
+++ b/packages/web/lib/db/databasemanager.class.php
@@ -31,6 +31,19 @@ class DatabaseManager extends FOGCore
     public static function establish()
     {
         /**
+         * Certain scripts don't use the database at all
+         * so we skip connecting to the DB entirely for those.
+         */
+        $noDBpattern = array(
+            'status\/bandwidth\.php$',
+            'status\/getfiles\.php$',
+            'status\/getservertime\.php$'
+        );
+        $noDBpattern = '#'.implode($noDBpattern, "|").'#';
+        if (preg_match($noDBpattern, self::$scriptname)) {
+            return;
+        }
+        /**
          * If the db is already connected,
          * return immediately.
          */
--- a/packages/web/lib/db/databasemanager.class.php
+++ b/packages/web/lib/db/databasemanager.class.php
@@ -41,7 +41,8 @@ class DatabaseManager extends FOGCore
             'status\/gethash\.php$',
             'status\/getservertime\.php$',
             'status\/getsize\.php$',
-            'status\/hw\.php$'
+            'status\/hw\.php$',
+            'status\/newtoken\.php$'
         );
         $noDBpattern = '#'.implode($noDBpattern, "|").'#';
         if (preg_match($noDBpattern, self::$scriptname)) {
--- a/packages/web/lib/db/databasemanager.class.php
+++ b/packages/web/lib/db/databasemanager.class.php
@@ -36,8 +36,12 @@ class DatabaseManager extends FOGCore
          */
         $noDBpattern = array(
             'status\/bandwidth\.php$',
+            'status\/freespace\.php$',
             'status\/getfiles\.php$',
-            'status\/getservertime\.php$'
+            'status\/gethash\.php$',
+            'status\/getservertime\.php$',
+            'status\/getsize\.php$',
+            'status\/hw\.php$'
         );
         $noDBpattern = '#'.implode($noDBpattern, "|").'#';
         if (preg_match($noDBpattern, self::$scriptname)) {
--- a/packages/web/lib/db/pdodb.class.php
+++ b/packages/web/lib/db/pdodb.class.php
@@ -139,10 +139,10 @@ class PDODB extends DatabaseManager
      * @throws PDOException
      * @return object
      */
-    private function _connect($dbexists = true, $always = false)
+    private function _connect($dbexists = true)
     {
         try {
-            if (self::$_link && !$always) {
+            if (self::$_link) {
                 return $this;
             }
             $type = DATABASE_TYPE;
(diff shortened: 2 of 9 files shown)


Sebastian-Roth:
Talked to Tom in chat. We are fine to merge this in and let people test.

packages/web/lib/fog/fogbase.class.php
```diff
@@ -1196,7 +1196,11 @@ protected static function validDate($date, $format = '')
         if (!$format) {
             $format = 'm/d/Y';
         }
-        $tz = new DateTimeZone(self::$TimeZone);
```
mastacontrola: We can greatly simplify this. Why have an if/else for a string?

Instead of the below we can do:
```
if (empty(self::$TimeZone)) {
    self::$TimeZone = 'UTC';
}
$tz = new DateTimeZone(self::$TimeZone);
```

packages/web/lib/fog/fogbase.class.php
```diff
@@ -2235,12 +2239,6 @@ protected static function getIPAddress()
                 $retVal
             );
         }
-        $test = self::$FOGURLRequests->isAvailable('http://ipinfo.io/ip');
```
mastacontrola: Rather than remove all the lines, I simply commented them. This way it's there if we need to re-add it for some reason :)

packages/web/lib/db/databasemanager.class.php
```diff
@@ -30,6 +30,24 @@ class DatabaseManager extends FOGCore
      */
     public static function establish()
     {
+        /**
```
mastacontrola: I LOVE THIS!

packages/web/lib/db/pdodb.class.php
```diff
@@ -543,9 +543,32 @@ public function dbName()
...
+        }
+        $this->_connect(true);
         return self::$_link;
     }
+    /**
+     * Returns the DB Link object
+     *
+     * @return boolean
+     */
+    public function ping()
```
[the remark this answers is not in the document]
Sebastian-Roth: You are totally right, my implementation of the ping function wasn't that great. But my intention was to make absolutely sure that we set `self::$_link` to false as soon as the query does not return properly.

The way you have it now I can imagine there  being cases where query would not trigger an exception but still fail/return false. Then we would not set `self::$_link` to false and might be stuck in that situation. What do you think about this:
```
public function ping()
{
    try {
        if (self::$_link) {
            if (self::$_link->query('SELECT 1')) {
                return true;
            }
        }
    } catch (PDOException $e) {
    }
    return (self::$_link = false);
}
```
Or simply toss out the whole try/catch altogether. Not sure if we need that. PDO connections don't seem to reliably trigger exceptions in my testing.

packages/web/lib/fog/fogbase.class.php
Sebastian-Roth: Nice!!

packages/web/lib/fog/fogbase.class.php
Sebastian-Roth: We have that code in the git history anyway so don't actually need it commented right here in place. But if you like, I am fine with commented too.

packages/web/lib/db/pdodb.class.php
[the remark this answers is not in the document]
Sebastian-Roth: Sure you are right about the try/catch stuff. It will always run the stuff that is outside of those.

But there could be cases where `self::$_link->query('SELECT 1')` does **not** trigger an exception **but** returns false. In that case we still have `self::$_link` set with the DB resource pointer.
1. I am not sure if that case can ever happen
2. I am not sure if it's a problem on the following query
3. Maybe I am wrong here but I'd want to reconnect to the DB in case "SELECT 1" returns false, what do you think?

Why should we optimize the code to the point where situations can occur that we don't know about.

Issue #284, opened 2019-01-25
Question: which is the right hook to step in during a client poll?


sctt:
Hello,

I'm coding a plugin which has to take decisions based on hosts IP addresses.
As soon as a host appears to be connected to the fog server, the plugin has to get its IP and do the work.

My idea was to take advantage of the polling system by triggering the plugin on every poll request, that's why I decided to use the HOST_INFO_EXPOSE hook, but then I found out that this hook had no longer been used so it would never be processed (is this hook deprecated? why it has been removed?).

Do you have any suggestion on which hook should I use to get this job done? I didn't find anything suitable... 
Should i just restore the HOST_INFO_EXPOSE hook by adding the trigger on the newer requestClientInfo function inside the fogpage class? This seems to me the easier way but maybe I'm missing something here, that's why i'm asking for your opinion about this.

Thank you in advance for your suggestions.

sctt:
thanks @Sebastian-Roth 

does the HOST_CHECKIN event trigger even when the host has not any scheduled task?

the hook/event i need is very generic that's why i couldn't be much specific.
let's say the plugin has to know the host IP address as soon as it appears to be **online** (cause it did a poll request, a checkin or anything else, but whatever it's going to be it has to be reliable: if that hook/event is triggered, the plugin is sure the host is online and the IP is available to do the work, and, on the contrary, if it's not triggered, the plugin is sure the host doesn't appear to be online so it has not to do the work)

if HOST_CHECKIN triggers only if there is a scheduled task, that's not what i need because it's too specific, i need something that will always get triggered when the host asks anything to the fog server, no matter of what it is or what's the outcome of that request.

if there's no hook/event suitable i'm still thinking about restoring that HOST_INFO_EXPOSE trigger... 
i hope i made myself clearer this time, thank you for your support.

Sebastian-Roth:
@sctt Yes, difinitely getting a better picture of what you ask here. AFAIK there is no such hook/event for what you are looking for.

> if there's no hook/event suitable i'm still thinking about restoring that HOST_INFO_EXPOSE trigger...

That hook won't help you at all. It's only triggered on scheduled tasks as well. I can't see of any way you could make it so it does exactly what you are trying to do.

I guess you just need to add an event or hook yourself or possibly find one looking through the code and testing. I am still not sure of the best place to put this. Most probably you will need to use at least two different hooks/events.

[...]

PR #289, opened 2019-02-05
SubnetGroups Plugin

sctt:
Adds the capability to **automatically** allocate host groups according to their IP subnet.
This is useful in those situations where (for example) IP subnets are associated to physical areas or rooms which are categorized using Fog groups.
With this plugin users are not required to **manually** update group mappings every time hosts change area or room (in our environment with hundreds of moving computers this is a real problem).

To solve this problem the plugin dynamically allocate groups by periodically evaluating hosts IP during their requestClientInfo request to the server, for this reason it has been added a new REQUEST_CLIENT_INFO hook, as referenced in issue #284.

The plugin has been tested on our system for a week and it works pretty well, but we are not fog experts, so we are interested in knowing if we missed something or did anything wrong with our coding by using your framework.

### DIFF
--- a/packages/web/lib/fog/fogpage.class.php
+++ b/packages/web/lib/fog/fogpage.class.php
@@ -3107,6 +3107,15 @@ abstract class FOGPage extends FOGBase
                 unset($key);
             }
             //echo json_encode($array, JSON_UNESCAPED_UNICODE);
+
+            self::$HookManager->processEvent(
+                'REQUEST_CLIENT_INFO',
+                 array(
+                     'repFields' => &$array,
+                     'Host' => self::$Host
+                 )
+            );
+
             $this->sendData(
                 json_encode(
                     $array,
--- a/packages/web/lib/plugins/subnetgroup/class/subnetgroup.class.php
+++ b/packages/web/lib/plugins/subnetgroup/class/subnetgroup.class.php
@@ -1,10 +1,10 @@
 <?php
 /**
- * Subnetgroups Class handler.
+ * Subnetgroup Class handler.
  *
  * PHP version 5
  *
- * @category Subnetgroups
+ * @category Subnetgroup
  * @package  FOGProject
  * @author   Tom Elliott <tommygunsster@gmail.com>
  * @author   sctt <none@none>
@@ -12,25 +12,25 @@
  * @link     https://fogproject.org
  */
 /**
- * Subnetgroups Class handler.
+ * Subnetgroup Class handler.
  *
- * @category Subnetgroups
+ * @category Subnetgroup
  * @package  FOGProject
  * @author   Tom Elliott <tommygunsster@gmail.com>
  * @author   sctt <none@none>
  * @license  http://opensource.org/licenses/gpl-3.0 GPLv3
  * @link     https://fogproject.org
  */
-class Subnetgroups extends FOGController
+class Subnetgroup extends FOGController
 {
     /**
-     * The subnetgroups table
+     * The subnetgroup table
      *
      * @var string
      */
-    protected $databaseTable = 'subnetgroups';
+    protected $databaseTable = 'subnetgroup';
     /**
-     * The subnetgroups fields and common names
+     * The subnetgroup fields and common names
      *
      * @var array
      */
--- a/packages/web/lib/plugins/subnetgroup/class/subnetgroup.class.php
+++ b/packages/web/lib/plugins/subnetgroup/class/subnetgroup.class.php
@@ -69,21 +69,4 @@ class Subnetgroup extends FOGController
             'group'
         ),
     );
-
-    /**
-     * Load the group object
-     *
-     * @return object
-     */
-    protected function loadGroup()
-    {
-        $group = self::getSubObjectIDs(
-            'Group',
-            array('id' => $this->get('groupID')),
-            'data'
-        );
-        if (isset($group[0])) {
-            $this->set('group', $group[0]);
-        }
-    }
 }
--- a/packages/web/lib/plugins/subnetgroup/class/subnetgroup.class.php
+++ b/packages/web/lib/plugins/subnetgroup/class/subnetgroup.class.php
@@ -69,6 +69,7 @@ class Subnetgroup extends FOGController
             'group'
         ),
     );
+
     /**
      * Load the group object
      *
--- a/packages/web/lib/plugins/subnetgroup/class/subnetgroup.class.php
+++ b/packages/web/lib/plugins/subnetgroup/class/subnetgroup.class.php
@@ -76,19 +76,13 @@ class Subnetgroup extends FOGController
      */
     protected function loadGroup()
     {
-        Route::listem(
-            'group',
-            'name',
-            false,
-            array('id' => $this->get('groupID'))
+        $group = self::getSubObjectIDs(
+            'Group',
+            array('id' => $this->get('groupID')),
+            'data'
         );
-
-        $Group = json_decode(
-            Route::getData()
-        );
-
-        if (isset($Group->groups[0])) {
-            $this->set('group', $Group->groups[0]);
+        if (isset($group[0])) {
+            $this->set('group', $group[0]);
         }
     }
 }
--- a/packages/web/lib/plugins/subnetgroup/class/subnetgroupmanager.class.php
+++ b/packages/web/lib/plugins/subnetgroup/class/subnetgroupmanager.class.php
@@ -1,10 +1,10 @@
 <?php
 /**
- * Manager class for subnetgroups
+ * Manager class for subnetgroup
  *
  * PHP Version 5
  *
- * @category SubnetgroupsManager
+ * @category SubnetgroupManager
  * @package  FOGProject
  * @author   Tom Elliott <tommygunsster@gmail.com>
  * @author   sctt <none@none>
@@ -12,23 +12,23 @@
  * @link     https://fogproject.org
  */
 /**
- * Manager class for subnetgroups
+ * Manager class for subnetgroup
  *
- * @category SubnetgroupsManager
+ * @category SubnetgroupManager
  * @package  FOGProject
  * @author   Tom Elliott <tommygunsster@gmail.com>
  * @author   sctt <none@none>
  * @license  http://opensource.org/licenses/gpl-3.0 GPLv3
  * @link     https://fogproject.org
  */
-class SubnetgroupsManager extends FOGManagerController
+class SubnetgroupManager extends FOGManagerController
 {
     /**
      * The base table name.
      *
      * @var string
      */
-    public $tablename = 'subnetgroups';
+    public $tablename = 'subnetgroup';
     /**
      * Perform the database and plugin installation
      *
--- a/packages/web/lib/plugins/subnetgroup/config/plugin.config.php
+++ b/packages/web/lib/plugins/subnetgroup/config/plugin.config.php
@@ -22,7 +22,7 @@
  * @link     https://fogproject.org
  */
 $fog_plugin = array();
-$fog_plugin['name'] = 'subnetgroups';
+$fog_plugin['name'] = 'subnetgroup';
 $fog_plugin['description'] = 'Associates host groups with IP subnets'
     . ' in order to automatically assign hosts according to their IP address';
 $fog_plugin['menuicon'] = 'fa fa-wifi fa-fw';
--- a/packages/web/lib/plugins/subnetgroup/hooks/addsubnetgroupapi.hook.php
+++ b/packages/web/lib/plugins/subnetgroup/hooks/addsubnetgroupapi.hook.php
@@ -1,10 +1,10 @@
 <?php
 /**
- * Injects subnetgroups stuff into the api system.
+ * Injects subnetgroup stuff into the api system.
  *
  * PHP version 5
  *
- * @category AddSubnetGroupsAPI
+ * @category AddSubnetGroupAPI
  * @package  FOGProject
  * @author   Tom Elliott <tommygunsster@gmail.com>
  * @author   sctt <none@none>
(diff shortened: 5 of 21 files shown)


Sebastian-Roth:
@sctt Thanks for your pull request! Great to see you got this worked out and you are happy to share your work with the public.

I just re-based your pull request to be merged into our `dev-branch` which currently work on.

Sebastian-Roth:
@sctt I don't have enough time to review the whole code right now but here are a few first comments that came to mind when I quickly had a look:
* You might want to change the `@author` information to add your name and e-mail
* We use `php-cs-fixer fix ${path_to_packages_web_dir} --rules=@PSR2` to normalize code style before pushing the code - [see here for info and install instructions](https://github.com/FriendsOfPHP/PHP-CS-Fixer)
* Instead of `getSubObjectIDs` it would be great to use the FOG API - I know we still have it in the code but we try to move towards the API as much as possible

You might just push more changes to your own master branch and they should add to this.

sctt:
@Sebastian-Roth thank you very much for helping in improving the plugin

i'm confused on how to fix your third point, most of plugins are actually using `getSubObjectIDs` quite abundantly, can you please point me out a piece of code that uses the newer style you are referring to?
as far as it can totally replace `getSubObjectIDs` i guess it's not a problem to comply with it.

Sebastian-Roth:
@sctt Here is an example on using the API: https://github.com/FOGProject/fogproject/blob/dev-branch/packages/web/lib/pages/fogconfigurationpage.class.php#L118 (line 118 to 124)

You are totally right that plugins still mostly (if not all) use the old style. But while you are at it it's definitely the best point to move to using the API calls straight away.

We have different internal API calls. Find those defined here: https://github.com/FOGProject/fogproject/blob/dev-branch/packages/web/lib/router/route.class.php#L230 and https://github.com/FOGProject/fogproject/blob/dev-branch/packages/web/lib/router/route.class.php#L429

mastacontrola:
Also, getSubObjectIDs does not exist, any more, in `working-1.6` branch.

I'll likely take a look at your plugin and make it work for the 1.6 versioning too, as well as cleanup a few things here and there.

While 1.5.5 has a lot of method usages and interweaved API callers, 1.6 is almost entirely independent when it comes to plugins.

I know this is a lot of information to take in, and I appreciate your effort and openness to help the project.

sctt:
@Sebastian-Roth i've removed those deprecated calls in favor of that `Route::listem` as you indicated.

@mastacontrola as it seems you are interested in releasing this, you must be informed that there are a series of doubts and imperfections which arose during the development that, even though are not of a concern in our specific environment, might be in different production ones.
i'll take care of posting here the list of the potential issues i'm aware of as soon as possible.

@Sebastian-Roth, @mastacontrola : please don't mind to point out other activities you want me to do on this plugin, yet i can't give you fast answers, i have enough time availability to work on this at least a couple of hours per week.

Sebastian-Roth:
@sctt Sorry for my late answer. I have a very busy week and won't get to this before the weekend. Just so you know it's not lost.

Sebastian-Roth:
@sctt Ok, here we go. Got a bit more time to look it over and play with it. While I have not done a full blown test (setting up this scenario would be way too much work) I have looked through the code again, installed it a my test server and played with the settings. Looked fine at first but then I ran into an issue. Adding a subnetgroup worked fine but when I click "List all SubnetGroups" I get a blank page an the following error in the logs:
```
[17-Feb-2019 19:42:59 UTC] PHP Warning:  Illegal string offset 'subnetgroupss' in /var/www/html/fog/lib/router/route.class.php on line 482
[17-Feb-2019 19:42:59 UTC] PHP Fatal error:  Uncaught Error: Cannot use string offset as an array in /var/www/html/fog/lib/router/route.class.php:482
Stack trace:
#0 /var/www/html/fog/lib/fog/fogpage.class.php(498): Route::listem(Object(Subnetgroups))
#1 /var/www/html/fog/lib/fog/fogpagemanager.class.php(220): FOGPage->index()
#2 /var/www/html/fog/management/index.php(58): FOGPageManager->render()
#3 {main}
  thrown in /var/www/html/fog/lib/router/route.class.php on line 482
```
From what I know about the framework (I am not the author of it) it generally uses singular terms for the classes and at some point does some magic appending 's' to that. See if you can replicate the issue and try to move to singular naming of all the classes (filenames as well as class definitions...).

sctt:
@Sebastian-Roth I'm not able to replicate your issue on our Fog v1.5.4, but i've renamed all classes and references to singular as you asked, please let me know if you still get those php errors during CRUD.

since you mentioned the complexity of setting up an appropriate scenario, i must say you don't really need your network to be organized into multiple subnets to test the behavior of the plugin.
let's suppose you have just a single network (like 192.168.1.0/24) with a couple of devices (one 192.168.1.100 and the other 192.168.1.150), then you can set a subnetgroup to 192.168.1.96/27 and another to 192.168.1.128/27.
if everything works we expect the first device to be assigned to the first subnetgroup, and the second device to the second one.

i hope that with this example you got how easy it is to test the plugin behavior just by using narrow subnets definitions on a single network.

Sebastian-Roth:
@sctt Thanks for updating the code to singular. I just tried your updated version and I am sorry to say that I was probably on the wrong track. Got the exact same error again. Only this time saying `PHP Warning:  Illegal string offset 'subnetgroups' in ...` (so no double s).

Looking a bit deeper it seems like something is clearing the `self::$data` Array in the [route.class.php code](https://github.com/sctt/fogproject/blob/master/packages/web/lib/router/route.class.php#L460). Before the `foreach` loop it's set to
```
Array
(
    [count] => 0
    [storagenodes] => Array
        (
        )

)
```
Seems perfectly fine but in the loop it's cleared (unset) and therefore the error `Cannot use string offset as an array ...`.

My guess is that PHP versions handle things differently here. I run PHP 7.2 on my test system. Which version of PHP do you have?

sctt:
@Sebastian-Roth we're running php 7.0 and fog 1.5.4
looks weird it doesn't work on your system that it's not so different than ours, consider that the whole plugin structure is based on the last version of that oldie wolbroadcast plugin (`getSubObjectIDs` to `listem` converstion apart), maybe you should try also to play with that wolbroadcast too and see if you have the same sort of issue.
as the only main difference is the introduction of `listem` maybe you should try my first commit version instead.
but as it seems a framework related problem i don't think i know enough about it's workings to tell you exactly what's going wrong on your system, let me know if i can be more helpful somehow.

Sebastian-Roth:
@sctt I am fairly sure this issue arises with PHP 7.1 and later versions. I am just trying to foresee any issues that could come up and try to get those fixed beforehand. You are definitely right about it being related to using the API (`listem`) and I hoped you might be interested to help us find out what's causing this.

As @mastacontrola just picked up on this we might get an answer from him. He knows the framework a whole lot better than I do!

sctt:
@Sebastian-Roth we created a test system with php 7.1 to check this out.
we understood the crash was caused by [this method](https://github.com/sctt/fogproject/blob/master/packages/web/lib/plugins/subnetgroup/class/subnetgroup.class.php#L77), so we rolled back to `getSubObjectIDs` just for this part and now everything works correctly, both in php 7.0 and 7.1.

as we observed, plugins like accesscontrol use `listem` in many parts, but still `getSubObjectIDs` for the "load" methods, that makes us think that maybe the API system is not ready yet to totally replace the older methods, at least for plugin interactions, but we might be wrong.

i hope @mastacontrola will directly work on this, cause this matter is not really related to this PR but more instead to your roadmap of enhancements from 1.5 to 1.6, and our plugin is based on 1.5 ones.

mastacontrola:
@sctt You can use the `Route::ids` method to get ID's only.

Here's the method from `dev-branch`:

`public function ids($class, $whereItems = [], $getField = 'id')`

It will return only the id's of the getField element.

whereItems is ['commonName' => <array|item>] (Common being what the left side of the databaseFields for each of the classes.)

Hopefully that helps.

mastacontrola:
So:

```
    protected function loadGroup()
    {
        $group = self::getSubObjectIDs(
            'Group',
            array('id' => $this->get('groupID')),
            'data'
        );
        if (isset($group[0])) {
            $this->set('group', $group[0]);
        }
    }
```
Could become:
```
    protected function loadGroup()
    {
        Route::ids('group', ['id' => $this->get('groupID')]);
        $groups = json_decode(Route::getData(), true);
        $this->set('group', array_shift($groupID));
    }
```

mastacontrola:
I might add, however, that the only reason you're doing this, I'm assuming, is because you can't necessarily trust that 'groupID' is containing a valid group Identifier?

sctt:
@mastacontrola the reason `loadGroup` is coded like that is because (like with most of the code) we took inspiration from other plugins.
we thought (maybe wrongly) that since the group is a reference, it needs to be loaded as other plugins do, like [this](https://github.com/FOGProject/fogproject/blob/master/packages/web/lib/plugins/location/class/location.class.php#L169), [this](https://github.com/FOGProject/fogproject/blob/master/packages/web/lib/plugins/windowskey/class/windowskey.class.php#L126) and [this](https://github.com/FOGProject/fogproject/blob/master/packages/web/lib/plugins/accesscontrol/class/accesscontrol.class.php#L140) example.

as you can see we merely worked on a current plugin structure without taking much initiative, the business logic is [here](https://github.com/FOGProject/fogproject/blob/working-1.6/packages/web/lib/plugins/subnetgroup/hooks/addsubnetgrouphost.hook.php#L76), the rest was done by watching other plugins (wolbroadcast mostly).

that means that most of your questions related to CRUD and not to business logic will almost ever have the same answer: other plugins do that.
for this reason please feel free to change anything you wish about framework dirty works, you sure know much more than us who are not even sure if `loadGroup` function is actually needed or not in this context (our code works even by commenting it), we left it cause we thought it might be needed in certain situations we are not aware of.

probably we are just using it the wrong way and we will learn by looking at your changes, thank you for your work.

sctt:
@mastacontrola speaking about business logic, i took a deeper look to your changes [here](https://github.com/FOGProject/fogproject/blob/working-1.6/packages/web/lib/plugins/subnetgroup/hooks/addsubnetgrouphost.hook.php#L76), and i can see a problem.

you decided to `$Host->removeGroup($SNGroup->groupID)->save();` inside the main loop [here](https://github.com/FOGProject/fogproject/blob/working-1.6/packages/web/lib/plugins/subnetgroup/hooks/addsubnetgrouphost.hook.php#L109), but since a group can be associated with more than a single subnetgroup, you might be going to remove a host from a group that has been just allocated by a previous subnetgroup in the same loop.

consider this example of subnetgroups config:
SG1     |     192.168.1.0/24    |     Classroom1
SG2     |     192.168.2.0/24    |     Classroom1

running your code with this config a host that matches 192.168.1.0/24 will never get assigned to Classroom1, that's because SG2 deletes the association that SG1 had previously defined.

to solve this you need to restore the structure of the original PR or to introduce a new constraint: you can't assign the same group to more than one subnetgroup.
i would choose the first solution cause i usually tend to give users as much flexibility as possible, in this case they would be free to decide if to use a comma separated list of subnets or just multiple subnetgroup definitions when they have more than a subnet for the same group.
but you might prefer the other way, you choose.

sctt:
`        // If resolve hostname returns a valid IP, set IP appropriately.
        // Otherwise, if the name is valid, use it.
        // Otherwise, return if base $ip is false.`

mastacontrola:
@sctt, What I might suggest then is making the group a unique item.

For me, it doesn't make sense to have two separate Subnet Groups that associate to the same Group ID. Logically speaking, just append your Subnet to the Subnets listing of the Subnet group you're working with.

For example, looping the same information twice, once to remove groups, and once to re-add doesn't makes sense.

Unless there's a particular reason you'd want to create two separate subnet elements that tie into the same exact group ID, I think this is just extra work.

sctt:
@mastacontrola we comply with your decision, as this is not the right place to start a philosophical discussion about coding principles on modern programming.

i saw you made the right checks (like for example [here](https://github.com/FOGProject/fogproject/blob/working-1.6/packages/web/lib/plugins/subnetgroup/pages/subnetgroupmanagement.page.php#L282) so there won't be any issues with the dynamic group allocation logic.

thank you again for the nice improvements you are applying to the plugin, don't hesitate to involve us if you need anything.

mastacontrola:
@sctt 
I totally understand what the thought process was. I just don't find that it makes much sense.  As you're dealing with subnet groups, the only reason I can foresee you wanting to separate is specifically so you can give the vlan names (or subnet names if you will.)  However, as you're dealing with sets of subnet's it seems a bit gratuitous to allow both separate vlan's matching the same group AND a comma separated list.  The list helps ensure things are grouped together, which ultimately is the point of the plugin to begin with.

I know some will prefer wanting to separate the names more implicitly while still assigning to the same group. I imagine, however, that a majority of people who would want to use this plugin will use it to designate specific subnets to specific groups.

For example, a school district might have a VLAN for student machines, a VLAN for teacher machines, and a VLAN for Computer Labs.  In these cases, I imagine one would not want the same group for student machines AND computer labs. 

Of course that's just one example to the seemingly infinite amount of possibilities.

Hopefully this all makes sense.

If you need help integrating your plugin into 1.5.x, please let me know.

Sebastian-Roth:
@sctt Hope you don't think we just keep pushing this along. I am more than happy to merge your pull request. Will do so if you are fine with it too. Just let me know.

Unfortunately I don't seem to find more time to join the detailed discussion you and Tom have.

What Tom is referring to about "integrating your plugin into 1.5.x" is that he's mainly working on a branch to push for a new 1.6 version of FOG where he's added your code already. While I'll merge the plugin code to be part of current FOG 1.5.x...

sctt:
@mastacontrola your point is very clear and we are not against the decision, it's fine and we consider ourselves satisfied with the work you did, no need to extend this discussion further considering that we are talking about such minor and irrelevant feature.

@Sebastian-Roth i'm not sure why you got the feeling we might have been somehow upset for anything about this PR, we are very happy about your interest in merging the plugin.

returning to 1.5.x work, as you can see we tried to fix the crash bug you reported by restoring `getSubObjectIDs` [here](https://github.com/sctt/fogproject/blob/master/packages/web/lib/plugins/subnetgroup/class/subnetgroup.class.php#L77).
probably not the best solution, we are not even sure that method is actually needed, please ask @mastacontrola about this matter.
for the rest, the plugin is working fine on our 1.5.x and in our view it might be considered ready to merge, but we are perfectly aware that more checks are still needed,  when you'll have enough time to perform deeper tests we'll still be available to try to help you in identifying issues and correcting bugs.
finally, since @mastacontrola changed the group allocation policy a little bit, our next commit is going to address these differences between 1.5 and 1.6 logic.

sctt:
@Sebastian-Roth We updated 1.5.x version to comply with latest @mastacontrola changes on plugin logic.
We removed loadGroup method to solve that php crashbug your reported, in our opinion that method it's currently useless and it's safe to be removed, you will tell us if it might be needed somehow.

please let us know what else do you need us to do before merging to 1.5.x

Sebastian-Roth:
@sctt Great, thanks again for your work and adding this feature plugin to the FOG project!

I did a test and didn't find any issues. All good to go. Merging now.

PR #295, opened 2019-03-12
Adding the remit-hello-interval option in the multicast command

fernandoGietz:
Create a little feature to add the possibility to config the remit-hello-interval in the udp-sender commands.
NOTE: require alter the nfsGroupMembers table in the database

### DIFF
--- a/packages/web/lib/pages/storagemanagementpage.class.php
+++ b/packages/web/lib/pages/storagemanagementpage.class.php
@@ -412,7 +412,7 @@ class StorageManagementPage extends FOGPage
             . '" autocomplete="off" class="form-control"/>'
             . '</div>',
             '<label for="bitrate">'
-            . _('Bitrates')
+            . _('Bitrate')
             . '</label>' => '<div class="input-group">'
             . '<input type="text" name="bitrate" id="bitrate" value="'
             . $bitrate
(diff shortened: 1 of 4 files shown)


packages/web/lib/pages/storagemanagementpage.class.php
```diff
@@ -411,12 +412,19 @@ public function addStorageNode()
             . '" autocomplete="off" class="form-control"/>'
             . '</div>',
             '<label for="bitrate">'
-            . _('Bitrate')
+            . _('Bitrates')
```
Sebastian-Roth: @fernandoGietz Did you change that label by intention?

fernandoGietz: No, a fat finger??

Sebastian-Roth:
Looks fine from my point of view. The change of schema.php will be a bit of an issue when merging into working-1.6 (@mastacontrola) but there is no way around it.

mastacontrola:
I think we can do the schema update relatively simply.

For working-1.6 I just add:
```
IF NOT EXISTS( SELECT NULL
            FROM INFORMATION_SCHEMA.COLUMNS
           WHERE table_name = 'nfsGroupMembers'
             AND table_schema = 'fog'
             AND column_name = 'ngmHelloInterval')  THEN
  ALTER TABLE `fog`.`nfsGroupMembers` ADD `ngmHelloInterval` VARCHAR(8) AFTER `ngmMaxBitrate`;
END IF;
```

mastacontrola:
Alright added this to working-1.6 with a testable approach. If the column doesn't already exist, create it. Otherwise just make it a null schema entry point.

PR #296, opened 2019-03-15
Added the identification of different OS when FOG does PingHost

fernandoGietz:
With these little changes FOG detects the OS and his status. We use the different error codes to know the OS.
Codes:
* 0 -> Windows
* 111 -> Linux and FOS (maybe MAC OSX)
* 110 or others -> Unknown (the computer can be shutdown or the firewall is up over the port 445)


Sebastian-Roth:
@fernandoGietz Looks good from my point of view. Thanks!!

PR #300, opened 2019-03-28
Added the feature to config the LDAP ports dinamically in the FOG Con…

fernandoGietz:
…figuration -> Plugin: LDAP


packages/web/lib/plugins/ldap/pages/ldapmanagementpage.class.php
```diff
@@ -205,7 +205,8 @@ public function add()
             $searchScope,
             true
         );
-        $ports = array(389, 636);
+//        $ports = array(389, 636);
```
mastacontrola: I'd highly recommend doing something like:

```
$ports = self::getSetting('LDAP_PORTS');
$ports = preg_replace('#\s+#', '', $ports);
$ports = explode(',', $ports);
```
This way people can enter the ports as `389,636` or `389, 636`.

Of course this isn't purely required, but just a nice thing I think most people will enter comma separated in the same fashion.

Also,

I'd recommend making the setting of the global value as a part of the ldap page menu items. This way you can change the ports from the ldap plugin page instead of having to fumble around and find out where it's located in the Global Settings. Again, a nice to have, but of course not fully required.

packages/web/lib/plugins/ldap/class/ldapmanager.class.php
```diff
@@ -143,7 +143,8 @@ public function install()
                 "INSERT INTO `%s`"
 		. " (settingKey,settingDesc,settingValue,settingCategory)"
 		. " VALUES"
-                . " ('FOG_USER_FILTER','Insert the uType codes comma separated. If you want to list all users, empty the textbox', '990,991','Plugin: LDAP')",
+                . " ('FOG_USER_FILTER','Insert the uType codes comma separated. If you want to list all users, empty the textbox', '990,991','Plugin: LDAP'),"
+                . " ('LDAP_PORTS','Insert the different ports comma separated.', '383,686','Plugin: LDAP')",
```
mastacontrola: You have a typo. Default LDAP ports should be 389,686.

packages/web/lib/plugins/ldap/pages/ldapmanagementpage.class.php
fernandoGietz: Done

packages/web/lib/plugins/ldap/class/ldapmanager.class.php
fernandoGietz: Fixed

packages/web/lib/plugins/ldap/pages/ldapmanagementpage.class.php
fernandoGietz: About the use of ldap page to config the plugin and not the FOG Configuration page, do we save the config setting in globalsettings table too?

packages/web/lib/plugins/ldap/pages/ldapmanagementpage.class.php
```diff
@@ -48,6 +48,13 @@ public function __construct($name = '')
         parent::__construct($name);
         global $id;
         global $sub;
+	$this->menu += array('PluginConfiguration' =>('Plugin Configuration'));
```
fernandoGietz: With this I add a new entry in the page menu. It works but I don't know if is the best way to do it.

packages/web/lib/plugins/ldap/pages/ldapmanagementpage.class.php
```diff
@@ -996,4 +1003,118 @@ public function editPost()
...
+        );
+        $ports = filter_input(
+            INPUT_POST,
+            'ports'
+        );
+	
+	try{
+		self::setSetting('FOG_USER_FILTER', $filter);
+                self::setSetting('LDAP_PORTS', $ports);
+		$msg = json_encode(
```
fernandoGietz: When I update the fields, PHP doesn't show the message :(

packages/web/lib/plugins/ldap/pages/ldapmanagementpage.class.php
mastacontrola: Shouldn't the line be:
`$this->menu['PluginConfiguration'] = _('Plugin Configuration');`

fernandoGietz: Added the suggested change, but when I go to the Plugin configuration page the entry in the menu section disappears O_o, If I go back to the List All LDAP server, appears again ...rare?

Sebastian-Roth:
@fernandoGietz Again sorry for the delay. Been on the road and lots of work piled up. Merging this now into the official repo.

@mastacontrola I will be merging this into `dev-branch` soon as well.

PR #301, opened 2019-03-28
Management Panel Changes

SamJakob:
This commit:

- Refactors `/management/other/index.php` to use PHP's 'alternative syntax for control structures.'
    - Read more: https://www.php.net/manual/en/control-structures.alternative-syntax.php
    - Why?
        - This improves code readability and comprehension for development.

- Displays a message when JavaScript is disabled using a `<noscript>` tag.
image
    (_Please, feel free to suggest style changes..._)
    - Why?
        - The FOG dashboard does not work correctly without JavaScript enabled. This makes it obvious to the user.

- Use AJAX for page loads
    - Why?
        - This saves data (particularly helpful when using cellular data) as less code needs to be sent.
       - This speeds up page load times; CSS and JS files that are common to every page remain loaded in the browser, plus AJAX and `interval`s, that would otherwise block the page from reloading, can be cancelled/aborted on page change. (I noticed load times about 50% faster.)
            - Feel free to try for yourself! You can toggle the setting on L789 (`AJAX_PAGE_LOADING_ENABLED`) in `/management/js/fog/fog.common.js` to enable/disable this feature.
        - This also improves UX; blank pages are frustrating!

- Isolates scripts that are common to every page into a seperate field (`Page->commonJavascripts`). This also makes not reloading these files on AJAX page loads easier.

- Includes bug fixes
    - Fixed a bug with user role association, (User management -> plugins -> user role) would return a generic error.


mastacontrola:
I like your suggestions and I'm going to add them, but I'm going to do so manually. This way it's following syntactical differences and consistency.

Hopefully you understand.

Thanks for the help, and please know you can do these things in the future.

packages/web/lib/fog/reportmaker.class.php
```diff
@@ -0,0 +1,338 @@
+<?php
```
mastacontrola: This class no longer exists. We don't use it anywhere in 1.6 in preference to generate reports for the items as needed directly within the respective pages.

SamJakob:
Sure thing! Thanks.

packages/web/lib/fog/reportmaker.class.php
SamJakob: Ahh. I thought there was something up with that. Is it done client-side in the page?

PR #306, opened 2019-04-16
Dev branch

fernandoGietz:
FIX added the ldapmanagement page that doesnt appear in the last merge


Sebastian-Roth:
@fernandoGietz Ok, seems fine from my point of view. Merging...

PR #309, opened 2019-05-08
FIX typo in location and site plugins

fernandoGietz:
When you update the location or site in the group page, this retrieve a blank page


Sebastian-Roth:
@fernandoGietz Sorry for the late reply. Have had a very busy week. Looks fine from my point of view. Though I don't have the time to test. Will merge anyway as I am sure this is fine.

PR #310, opened 2019-05-21
Dev branch

fernandoGietz:
Added the feature to queue multicast tasks when is not any free slot in the storage node.
I change the packages/web/lib/fog/storagegroup.class.php file to do well the sum of the queued tasks and active tasks in the dashboard. With this change one multicast task use one free slot in the storage node and not as many as compueter are in the session.


Sebastian-Roth:
@mastacontrola Tom what do you think of this?

fernandoGietz:
I need more time to do more tests XD I detect a bug

Sebastian-Roth:
@fernandoGietz Sorry but I have pushed a couple of PHP code style fixes that now produce conflicts with your edits on the same files. Can you possibly add 521aa5bd4 to your code as well so the merge will be fine again?

@mastacontrola Any chance you can take a look at this as well?

fernandoGietz:
I have fixed the problem when the service. Now works fine, I tested it in my test environment, I can ´t test it in multinode environment

Sebastian-Roth:
@fernandoGietz Still not exactly sure how to go about this. I am not able to properly test this but I see you have done some basic testing. Are you still around for discussing things here?

Would be great to get @mastacontrola to comment on this as well.

mastacontrola:
Sorry, meant to respond. I don't see any issues with having the "queue" show the individual multicast task when no slots are available.

Sebastian-Roth:
@fernandoGietz Thanks again for your pull request. I just merged it into the official `dev-branch`.

Please remember to run the code styling tool (`php-cs-fixer self-update ; php-cs-fixer fix packages/web/ --rules=@PSR2`) every now and then so we minimize commits that just fix the code styling, thanks!

PR #313, opened 2019-06-08
Add FOGBackup options

jdorel:
Add `--no-images`, `--no-snapins`, and `--no-reports`.

Also show the backup path and name

Also, the [documentation on the wiki](https://wiki.fogproject.org/wiki/index.php/Backing_up_FOG) appears to be wrong:
```
./FOGBackup.sh [directory]
```
Should be
```
./FOGBackup.sh -b [directory]
```

### DIFF
--- a/utils/FOGBackup/FOGBackup.sh
+++ b/utils/FOGBackup/FOGBackup.sh
@@ -25,13 +25,13 @@ while getopts "$optspec" o; do
                     backupPath=$OPTARG
                     ;;
                 no-reports)
-                    noBackupReports='true'
+                    noBackupReports=1
                     ;;
                 no-snapins)
-                    noBackupSnapins='true'
+                    noBackupSnapins=1
                     ;;
                 no-images)
-                    noBackupImages='true'
+                    noBackupImages=1
                     ;;
                 *)
                     if [[ $OPTERR -eq 1 && ${optspec:0:1} != : ]]; then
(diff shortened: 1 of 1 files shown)


Sebastian-Roth:
@jdorel Sorry for the long delay. Have been away and still trying to catch up with everything.

Thanks for sending in the pull request. I may re-base this to the `dev-branch` as we use this to work on things.

Sebastian-Roth:
@jdorel I usually tend to use numeric values for binary checks. Something like this:
```
noBackupReports=1
...
[[ "$noBackupReports" -ne 1 ]] && backupReports
```
What do you think? Sure it's literally the same so if you prefer to use "true" instead I can still go ahead and merge this. Otherwise you can just change the code, push to your repo again and it should reflect that in the existing pull request here before I merge it in.

Sebastian-Roth:
@jdorel I usually tend to use numeric values for binary checks. Something like this:
```
noBackupReports=1
...
[[ "$noBackupReports" -ne 1 ]] && backupReports
```
What do you think? Sure it's literally the same so if you prefer to use "true" instead I can still go ahead and merge this. Otherwise you can just change the code, push to your repo again and it should reflect that in the existing pull request here before I merge it in.

Sebastian-Roth:
@jdorel As well fixed the wiki, thanks for mentioning this! :+1:

jdorel:
@Sebastian-Roth I usually prefer to be more verbose and explicit, rather than having to think about negations and bash test commands (`-ne`), both for myself and for newcomers. That being said, I think it's best to use the current style.

Sebastian-Roth:
@jdorel Definitely a point you made there! Probably I've just been too deep into scripting and can't see things from a clear angle anymore. Thanks for updating the pull request though! Will merge that in now.

Issue #307, opened 2019-04-17
Add support for ARM machines


parkerberb-nfv:
I think we should fully support ARM client machines in FOG.

I have submitted some code related to this in the past: 
[here](https://github.com/FOGProject/fogproject/pull/239)
[and here](https://github.com/FOGProject/fos/pull/10)

I currently use FOG with the above changes to manage my ARM64 based machines, but this takes a few shortcuts. For ARM to be fully supported, we would need:

1) Support for 64 and 32 bit arm kernel and init images, just like the user can currently select with kernel and init to use for their 64 bit / 32 bit x86 machines

2) Reliable detection of system arch. Currently the default.ipxe script relies on cpuid, which is x86 specific: [docs](http://ipxe.org/cmd/cpuid). I fall back on [buildarch](https://ipxe.org/cfg/buildarch), but this is not a perfect solution either. This should identify the host as one of (32-bit x86, 64-bit x86, 32-bit ARM, 64-bit ARM) and use that to grab the correct kernel and init when pxe booting.

3) Bundling of ARM-based grub / refind / ipxe / kernel / initrd. My PR above allows the regular FOS system to build arm based kernel and init, but I am not familiar with how you consume / build everything and bundle it into a FOG release.

Sebastian-Roth:
@parkerberb-nfv Building the ARM inits I just ran into a question that caught me blank. The Buildroot config has many different choices of "Architecture Variants". In the old config we were using `arm926t`. Is this still correct for your machines? Here is the full list:
```
Target Architecture Variant
* armv4 cores
  1. arm920t (BR2_arm920t)
  2. arm922t (BR2_arm922t)
  3. fa526/626 (BR2_fa526)
  4. strongarm sa110/sa1100 (BR2_strongarm)
* armv5 cores
> 5. arm926t (BR2_arm926t)
  6. iwmmxt (BR2_iwmmxt)
  7. xscale (BR2_xscale)
* armv6 cores
  8. arm1136j-s (BR2_arm1136j_s)
  9. arm1136jf-s (BR2_arm1136jf_s)
  10. arm1176jz-s (BR2_arm1176jz_s)
  11. arm1176jzf-s (BR2_arm1176jzf_s)
  12. mpcore (BR2_arm11mpcore)
* armv7a cores
  13. cortex-A5 (BR2_cortex_a5)
  14. cortex-A7 (BR2_cortex_a7)
  15. cortex-A8 (BR2_cortex_a8)
[...]

PR #327, opened 2019-08-06
Fix Arm Boot 

parkerberb-nfv:
This pr adds the necessary binaries for ARM support.
The refind binary was downloaded from sourceforge: https://sourceforge.net/projects/refind/
The snp-arm.efi binary I believe I had to compile myself on an ARM-based machine. This correctly sets the ${buildarch} variable so that the default.ipxe script can correctly detect arm based machines. It also embeds the necessary ipxe script file.

I also updated the bootmenu.class.php file. For whatever reason, imgfetching the config file for refind broke everything on all the arm machines I tested.

With these two files, and the arm kernel / init, I am able to use fog perfectly normally on my ARM based machines. The only thing required from the user at this point is to manually set the kernel / initrd for the arm machines to point towards the arm-based kernel / init.


parkerberb-nfv:
Forgot to mention Issue #307

Sebastian-Roth:
@parkerberb-nfv Nice!! Just in time for the 1.5.8 release we are preparing. So this is going to be official very soon!

PR #333, opened 2019-10-02
Sanity checking remote file data

mjrider:
This should fix the issue that when a nas is used as storage node, that fog will continuously try to sync the data
example:
https://forums.fogproject.org/topic/11483/continuous-fog-storage-node-replication-problem/3


mastacontrola:
I've made the appropriate adjustments for working-1.6 for both of these. Will work on adding into the dev-branch code. I don't know if you can reset the pull request to goto this particular branch? (dev-branch) as opposed to the master branch?

Issue #319, opened 2019-06-20
Connection failure occurred while initializing database


swenye:
FOG Version:1.5.6
Branch:master
A new installation and with no empty mysql password. when at the step of init database it occurred connection failure. mysql instance is running.after i using mysqladmin change the password, it solved. My password include a "!" i don't know if it caused.

Sebastian-Roth:
@swenye Sorry for the delay. Have been away and just catching up with all the things that came up since then.

So you had a server with mysql already installed and mysql root passwort set already? Or did you start with a complete fresh server and just told the FOG installer that you don't want en empty password?

Just trying to get more details before I try to find and fix this bug. Sorry for that but I have very limited time at the moment.

swenye:
Hi, I start with a fresh server. Yes, I don't want empty password, and I input some a password like sdfab!   
take your time I just report it.

Sebastian-Roth:
@swenye I have had a first look at this and see why this is failing. As I am working on improving that mysql part of the FOG installer at the moment anyway I'll fix this as well. Will be a couple of days as it needs extensive testing...

Sebastian-Roth:
@swenye Can you please do me a favor before I fully add this to the code?

Please download the EFI binaries I extracted from the MemTest USB key image from their website here: [memtest_v8.zip](https://github.com/FOGProject/fogproject/files/3365955/memtest_v8.zip)

Put those files in `/var/www/html/fog/service/ipxe/` on your FOG server and add a new iPXE menu item for testing - FOG Configuration -> iPXE New Menu Entry:
* Menu Item: MemTest UEFI
* Description: Test new MemTest UEFI version
* Parameters: `chain ${boot-url}/service/ipxe/memtest_x64.efi || goto MENU`
* Menu Show with: All Hosts

This is more or less untested so it might need some adjustment. Call out if it doesn't work that way. If it does please let us know if you are able to properly boot the new memtest this way on UEFI hardware.

Sebastian-Roth:
@swenye Sorry for the long delay but finally I found the time to work in this. It's currently in testing stage and you are more than welcome to give it a try.

```
git clone https://github.com/FOGProject/fogproject
cd fogproject/
git checkout db-security
cd bin/
./installfog.sh
```

Sebastian-Roth:
This should be fixed in the above mentioned commit.

PR #355, opened 2019-12-19
Improve database security by using less privileged account and force users to set a DB root password

Sebastian-Roth:


Sebastian-Roth:
Fixes #319

PR #380, opened 2020-04-02
Fix two typos in ipxescripts

bchr:
I hope I did this correctly,
Thank you for fixing this.


Sebastian-Roth:
@bchr Yes, well done! And thanks again.

PR #405, opened 2020-09-28
Add FOG Analytics 

wayneworkman:
This adds the components needed for FOG Analytics.
The FOG Installer was modified to setup a weekly cronjob. This cron job is created such that a random day, hour, and minute is chosen during installation to create the cron job with. This job runs once per week. If it fails, it is of no consequence to the FOG Server's operation. The cron job collects the following information from the FOG Server:

- FOG Version
- OS Name
- OS Version

This information is then sent to the FOG Analytics API, which records these three things. This information will be used to inform the FOG Developers of what OSs and Versions are out there being used for FOG going into the future such that they can make more-informed decisions for the FOG Project.

After installed, there are some settings for this located at `/opt/fog/analytics/settings` which will allow you to choose the day, hour, and minute, the log file, and the user to run the job as. This file is used by the FOG installer to re-create the cron job upon each installation run.

Discussion is here: https://forums.fogproject.org/topic/14668/popularity-contest

### DIFF
--- a/bin/installfog.sh
+++ b/bin/installfog.sh
@@ -630,6 +630,7 @@ while [[ -z $blGo ]]; do
                     writeUpdateFile
                     linkOptFogDir
                     updateStorageNodeCredentials
+		    setupFogAnalytics
                     echo
                     echo " * Setup complete"
                     echo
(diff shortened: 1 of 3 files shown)


wayneworkman:
Discussion is here: https://forums.fogproject.org/topic/14668/popularity-contest

mastacontrola:
I've added the information so that we have all paths taken care of.

Thank you for the PR and all!

PR #414, opened 2020-11-09
Added web login logging

lukebarone:


lukebarone:
https://forums.fogproject.org/topic/13873/security-request-integrated-fail2ban-for-login-window?_=1604703751013 - Part of adding logging abilities to the Web Portal. Tested on Debian Buster, fresh install.

mastacontrola (commented):
While I understand what changes you're making, I don't fully understand why this needs to be stored in /var/log/fog and why it would need to be created by the fog installer?

As what you're looking for is regarding HTTP, I would imagine a better suitable option would be to log in the webroot kind of like how we log SCHEMA updates.

What do I mean by this?
Something like what you have for processlogin.class.php:
```
if (!self::$FOGUser->isValid()) {
    error_log(
        sprintf(
            '%s - %s - %s - %s: %s %s'
            $_SERVER['REMOTE_ADDR'],
            $_SERVER['HTTP_USER_AGENT'],
            _('Login failed'),
            _('username'),
            $uname,
            _('failed to login')
        ),
        3,
        BASEPATH . 'fog_login_failed.log'
    );
}
```

And a similar sequence but changing "successful" log with successful information.

I think it makes it more native to the application and would not require adding to the FOG Installer itself.

PR #415, opened 2021-01-03
Add User=fogproject to systemd unitfiles

KaiHerlemann:
I have added a "User=fogproject" line to all systemd unit files.
1) The daemons otherwise run with "root" rights. Running daemons, processes in general with root privileges is a security risk and should be avoided if possible.
It seems that in the past they were already running with fogproject ownership (although there was no line as suggested in the systemd unit files), as images were stored on our server with fogproject ownership. – But not always, sometimes in the past they had been saved with root ownership. This means that the responsible process then had run with root rights, as it does now.
2) If the images are stored with root ownership, it is not possible to delete them via the web interface. The images will then be deleted in the database, but not in the /images folder. You have to delete them manually.

We had the problem on CentOS 7.x, as well as now on Debian 10.7.

### DIFF
--- a/packages/systemd/FOGImageReplicator.service
+++ b/packages/systemd/FOGImageReplicator.service
@@ -22,6 +22,7 @@ Type=simple
 Restart=always
 RestartSec=1
 ExecStart=/usr/bin/env php /opt/fog/service/FOGImageReplicator/FOGImageReplicator
+User=fogproject
 
 [Install]
 WantedBy=multi-user.target
--- a/packages/systemd/FOGImageSize.service
+++ b/packages/systemd/FOGImageSize.service
@@ -21,6 +21,7 @@ Type=simple
 Restart=always
 RestartSec=1
 ExecStart=/usr/bin/env php /opt/fog/service/FOGImageSize/FOGImageSize
+User=fogproject
 
 [Install]
 WantedBy=multi-user.target
--- a/packages/systemd/FOGMulticastManager.service
+++ b/packages/systemd/FOGMulticastManager.service
@@ -23,6 +23,7 @@ Type=simple
 Restart=always
 RestartSec=1
 ExecStart=/usr/bin/env php /opt/fog/service/FOGMulticastManager/FOGMulticastManager
+User=fogproject
 
 [Install]
 WantedBy=multi-user.target
--- a/packages/systemd/FOGPingHosts.service
+++ b/packages/systemd/FOGPingHosts.service
@@ -21,6 +21,7 @@ Type=simple
 Restart=always
 RestartSec=1
 ExecStart=/usr/bin/env php /opt/fog/service/FOGPingHosts/FOGPingHosts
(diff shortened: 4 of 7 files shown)


George1422:
I've been working on and testing implementing NFSv4 for fog 1.6.x series. https://forums.fogproject.org/post/139585

For the NFS shares the FOG client connects as root (because that is what the process runs as on the capture / deploy engine). So it will always upload to the image store as the root user. I also see that as a security risk. So for NFSv4 I'm looking into changing the NFS share to all_squash from root_squash and then defining a user ID and group ID to save the files as. These should be set to the fogproject user. This will dovetail into the changes you've made nicely. 

all_squash,anonuid=1000,anongid=1000  where 1000 would be the extracted user id and group id of the fogproject user.

Sebastian-Roth:
@KaiHerlemann Thanks for bringing this up and your pull request on this. While I like the idea of running the services as non-privileged user I don't think that what you hope to get from this will actually be true. But before I get into the details of this I want to ask how well these changes have been tested yet? The services need to write log files and fork other processes to actually do the tasks and I am wondering if all of that is working out of the box as `fogproject` user. Don't get me wrong, I agree that we should move towards a more secure setup of FOG (when this project came to life many years ago security was not given too much thought as times were a bit different then) but we don't want to break things on the other hand. This change could have consequences in many of FOG's components (e.g. like replication to other nodes) and we need to make sure to test all of it beforehand.

> It seems that in the past they were already running with fogproject ownership (although there was no line as suggested in the systemd unit files), as images were stored on our server with fogproject ownership. – But not always, sometimes in the past they had been saved with root ownership. This means that the responsible process then had run with root rights, as it does now.

The FOG services don't play a role in how the ownership of images are being set because upload/capture always happen via NFS (as mentioned by @George1422) to folder `/images/dev/` and then will be renamed/moved to `/images/` using FTP. This part of FOG is many years old and we discuss changing it but it will take time.
What you see with some images owned by fogproject and others by root probably comes from the FOG installer running a simple `chown` command on the `/images/` directory. So when upgrading to FOG 1.5.9 it "chown"ed the existing images but those uploaded later are owned by root.

> If the images are stored with root ownership, it is not possible to delete them via the web interface. The images will then be deleted in the database, but not in the /images folder. You have to delete them manually.

There are two different ways you can delete images through the web UI:
1. When you delete an image from the list/search view by marking the checkbox and click the “Delete” button this will only remove the image definition from the database but won’t purge the files from disk. This is by intention because we want to prevent people from marking several images unintentionally and nuke those forever.
2. Go to the image settings in the web UI and click the "Delete" tab at the top. Here you can set a checkbox to actually purge the files from disk as well.

Sebastian-Roth:
@KaiHerlemann Just saw this pull request and was wondering if you'd comment on our replies?

Sebastian-Roth:
Gonna close this now. Let is know if you are keen to revive this topic.

PR #428, opened 2021-06-01
Update config.sh for Debian 11.x (bullseye)

smonaica:
* Debian Bullseye runs PHP 7.4
* php-gettext becomes php-php-gettext in Bullseye

### DIFF
--- a/lib/ubuntu/config.sh
+++ b/lib/ubuntu/config.sh
@@ -20,6 +20,7 @@
 [[ -z $packageQuery ]] && packageQuery="dpkg -l \$x | grep '^ii'"
 if [[ $linuxReleaseName == +(*[Bb][Ii][Aa][Nn]*) ]]; then
     sysvrcconf="sysv-rc-conf"
+    phpgettext="php-gettext"
     case $OSVersion in
         8)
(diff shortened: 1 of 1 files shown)


mastacontrola (changes requested):
Can you change the 11) section and see if it follows the same structure as Ubuntu 21 does?

Basically, php-gettext is a part of php7.4-common. So if you attempt to install php7.4-gettext, it will use the common package and remove the need to get php-gettext. php-php-gettext is an alternate package that I've not seen a requirement for. If php7.4-gettext doesn't automatically resolve to php7.4-common, then we will look at making these changes.

I've pushed what I hope will perform the same fix as what ubuntu did. Please see if that works, if so we can close this PR.

Thank you,

smonaica:
I just tried installing on bullseye from a checkout of `dev-branch`, and it worked.
So it looks like this pull request can be closed.
Can we expect a new stable release soon?

Sebastian-Roth:
Closing this as fixed in `dev-branch` ([reference](https://github.com/FOGProject/fogproject/issues/429#issuecomment-853527539)).

PR #452, opened 2022-01-05
take out class PXe function of dhcpd.conf of subnet for permit at lea…

cavazquez:
…st two subnet with diferent gw and next-server.


Sebastian-Roth:
@cavazquez Thanks for bringing this up. Having the class definitions within the subnet block is probably just a relict from old times. We just didn't bother about changing it as it has been working great for many years.

Citing Sten Carlsen's [post on the official mailinglist](https://lists.isc.org/pipermail/dhcp-users/2020-April/022040.html)
> [...] AFAIK class statements are global in nature [...]

Before we merge your pull request I need to ask you to apply some minor changes to it:
* We currently work on the branches `dev-branch` and `working-1.6`. Can you please prepare your pull request towards one of these branches instead of `master`. It doesn't matter which of the two you choose because we'll cherry-pick the commit to the other one later on.
* With your commit you add two "empty" lines to the dhcpd.conf file. While it's not a problem I still wonder why. If you really think you want to keep those then I ask you to make sure you use correct indention as well. Those two lines (partly) use tabulator indention instead of spaces. We try to keep indention the same (using spaces) in all places.

lib/common/functions.sh
```diff
@@ -2482,44 +2482,46 @@ configureDHCP() {
...
-            echo "        match if substring(option vendor-class-identifier, 0, 14) = \"AAPLBSDPC/i386\";" >> "$dhcptouse"
-            echo "        option dhcp-parameter-request-list 1,3,17,43,60;" >> "$dhcptouse"
-            echo "        if (option dhcp-message-type = 8) {" >> "$dhcptouse"
-            echo "            option vendor-class-identifier \"AAPLBSDPC\";" >> "$dhcptouse"
-            echo "            if (substring(option vendor-encapsulated-options, 0, 3) = 01:01:01) {" >> "$dhcptouse"
-            echo "                # BSDP List" >> "$dhcptouse"
-            echo "                option vendor-encapsulated-options 01:01:01:04:02:80:00:07:04:81:00:05:2a:09:0D:81:00:05:2a:08:69:50:58:45:2d:46:4f:47;" >> "$dhcptouse"
-            echo "                filename \"ipxe.efi\";" >> "$dhcptouse"
-            echo "            }" >> "$dhcptouse"
+	    echo " " >> "$dhcptouse"
```
Sebastian-Roth, line 2485: As mentioned in my comment you have a tabulator in front of the "echo" instead of just spaces. Please correct if you want to keep this empty line. Though I don't think this empty line makes sense within the subnet block.

lib/common/functions.sh
```diff
@@ -2482,44 +2482,46 @@ configureDHCP() {
...
+            echo "}" >> "$dhcptouse"
+	    echo " " >> "$dhcptouse"
```
Sebastian-Roth, line 2487: This empty line also uses a tabulator, please correct. Adding an "empty" line to the dhcpd.conf kind of makes sense at this place where the subnet block ends and classes start. But I still wonder why you add a space? Just `echo "" >> "$dhcptouse"` would be fine.

Sebastian-Roth (commented):
Just adding some more details to my comment.

Sebastian-Roth:
@cavazquez Thanks! Merging this now.

Sebastian-Roth:
@cavazquez This was an easy merge even though your changes were towards `master` branch. For further pull requests may I ask you to do them on `dev-branch` directly. Thanks again!

PR #455, opened 2022-01-31
Add ability to specify additional TFTP server command line options

blkeller:
This PR adds a new setting to `.fogsettings` called `tftpAdvOpts` (for "TFTP Advanced Options") that allows a user to specify additional command line arguments to pass to the TFTP server.  Currently, FOG hard-codes the command line options used to run the TFTP server, and any manual changes that are made to either `/etc/default/tftpd-hpa` or `/etc/xinetd.d/tftp` for these options are clobbered by the installer after each installation/upgrade run.  This new option provides a way to make persistent changes to these options that will survive a FOG upgrade or re-install.  The default value of `tftpAdvOpts` is an empty string, so the TFTP server command line options will not be altered at all unless a user intentionally modifies this setting in `.fogsettings`.

My use case that inspired this new setting is my need to pass `--blocksize 1388` as an extra command line option to the TFTP server.  Our organization PXE-boots some of our devices over a WireGuard tunnel that connects a remote site back to our FOG server at our main office.  The WireGuard tunnel is completely transparent to the client devices, except that the inside of the tunnel has a reduced MTU of 1420.  This is in order to accommodate the network overhead introduced by WireGuard's encapsulation and to prevent unnecessary packet fragmentation on the outside of the tunnel by giving devices inside of the tunnel the information they need to perform their own proper packet sizing or (failing that) packet fragmentation.

While this works for other more modern protocols and smarter devices, unfortunately, most PXE firmware is too simple and brain-stem-like to be able to reassemble fragmented IP packets, so it will silently drop the traffic instead.  Because of this, we have to fix the mismatch at the application level instead of the network level.  The TFTP maximum block size must be reduced in order to fit inside of the WireGuard tunnel's 1420 MTU without fragmenting.  Adding `--blocksize 1388` tells the TFTP server to properly size its application data to fit within the lowered 1420 MTU rather than to generate larger data blocks (and hence larger IP packets) that rely on proper IP fragmentation and reassembly at the network level.

### DIFF
--- a/bin/installfog.sh
+++ b/bin/installfog.sh
@@ -388,6 +388,7 @@ echo "Done"
 [[ -z $httpproto ]] && httpproto="http"
 [[ -z $armsupport ]] && armsupport=0
 [[ -z $mysqldbname ]] && mysqldbname="fog"
+[[ -z $tftpAdvOpts ]] && tftpAdvOpts=""
 [[ -z $fogpriorconfig ]] && fogpriorconfig="$fogprogramdir/.fogsettings"
 #clearScreen
 if [[ -z $* || $* != +(-h|-?|--help|--uninstall) ]]; then
--- a/lib/common/functions.sh
+++ b/lib/common/functions.sh
@@ -582,7 +582,7 @@ configureTFTPandPXE() {
                 if [[ -f /etc/systemd/system/fog-tftp.service ]]; then
                     mv -fv /etc/systemd/system/fog-tftp.service "/etc/systemd/system/fog-tftp.service.${timestamp}" >>$error_log 2>&1
                 fi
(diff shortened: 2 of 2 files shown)


George1422:
I agree that adding the block size option to the tftp server will add value. There has been several posts on the FOG Forums where FOG Admins have issues when imaging over the WAN and the MTU of the WAN is below the standard block size for the tftp server. This needs to be done for exactly what you mentioned, the pxe rom doesn't reassemble fragmented packets too well. When the target computer fails to download the iPXE boot loader the first question I typically ask is, is the FOG server and target computer on the same LAN? Unfortunately the FOG Admin doesn't realize that he/she needs to adjust the block size until after FOG is already installed. So at least post install updating the .fogsettings file will ensure that all subsequent version upgrades will detect this setting and apply it to the tftp server startup file.

blkeller:
I resolved the merge conflicts with the latest changes in `dev-branch`.  I made sure to apply the new setting from this PR to the systemd unit file that was just added in `dev-branch` as well.  Please let me know if there's anything else you need from me in order to consider this PR for merging.  Thanks!

Sebastian-Roth:
@blkeller Thanks for your pull request to add TFTP options and sorry for the delay.

Looks good to me, merging it now.

PR #460, opened 2022-02-20
Basic readability changes, and shortening of quite a few lines

knotapun:
Hello, these are rather simplistic changes, but seeing as this is more of a first commit, I think that's expected. 

The first set of changes is largely just find and replacing the string:
`${workingdir}/error_logs/fog_error_${version}.log` with `$error_log`
Simple, and perhaps silly, but I think it makes a readability difference.

Second, I checked the behavior of `lsb_release` and found that a considerable amount of usage of `awk` wasn't necessary,
Adding `-s` to the options used when retrieving individual attributes gives us an output without a header, which allows us to just directly set the variables.

I changed the function named `help()` to `usage()` which is similarly small, but bash has a built-in called `help`, and it's consistent with the `FOGBackup.sh`

Third, and final of the changes is moving a blanket failure case to the top, which is a test case on the result of `uname -s`. The only distinction between the behavior before and after this change is that 

Sorry that these aren't more substantive changes, but I'm making an earnest attempt to get a real commit in.

### DIFF
--- a/bin/installfog.sh
+++ b/bin/installfog.sh
@@ -37,7 +37,7 @@ if [[ ! $(echo "$OS" | tr [:upper:] [:lower:]) =~ *linux* ]]; then
     exit 2 # Fail OS Check
 fi 
 
-error_log=$/workingdir/error_logs/fog_error_${version}.log
+error_log=${workingdir}/error_logs/fog_error_${version}.log
 timestamp=$(date +%s)
 backupconfig=""
 . ../lib/common/functions.sh
--- a/bin/installfog.sh
+++ b/bin/installfog.sh
@@ -30,6 +30,7 @@ if [[ $? -eq 1 || $(echo $PATH | grep -o "sbin" | wc -l) -lt 2 ]]; then
     echo "of the su command as it is important to load root's environment)."
     exit 1
 fi
+error_log=$/workingdir/error_logs/fog_error_${version}.log
 timestamp=$(date +%s)
 backupconfig=""
 . ../lib/common/functions.sh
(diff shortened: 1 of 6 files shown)


Sebastian-Roth:
@knotapun Awesome, thanks for sending in this first PR!!

> The first set of changes is largely just find and replacing the string: `${workingdir}/error_logs/fog_error_${version}.log` with `$error_log` Simple, and perhaps silly, but I think it makes a readability difference.

Absolutely.

> Second, I checked the behavior of `lsb_release` and found that a considerable amount of usage of `awk` wasn't necessary, Adding `-s` to the options used when retrieving individual attributes gives us an output without a header, which allows us to just directly set the variables.

**Allow me one question here.** Did you check this on several distros? Not all but at least Ubuntu, Debian and one of the redhat-based ones?

> I changed the function named `help()` to `usage()` which is similarly small, but bash has a built-in called `help`, and it's consistent with the `FOGBackup.sh`

Good catch.

> Third, and final of the changes is moving a blanket failure case to the top, which is a test case on the result of `uname -s`. The only distinction between the behavior before and after this change is that

.... part of the sentence missing here? Anyhow, I think this is fine.

> Sorry that these aren't more substantive changes, but I'm making an earnest attempt to get a real commit in.

No worries, we really appreciate your work on this a lot!

knotapun:
> .... part of the sentence missing here? Anyhow, I think this is fine.
Ope, sorry, someone came in as I was typing that.
The only distinction between the behavior before and after this change is that the script won't attempt to parse command line parameters before exiting.

> Allow me one question here. Did you check this on several distros? Not all but at least Ubuntu, Debian and one of the redhat-based ones?
I've tested `lsb_release -s` on Pop! (Debian derivative), CentOS-7, Mint, and Ubuntu. I can test it on Debian as well, if you would like.

Sebastian-Roth:
> > Allow me one question here. Did you check this on several distros? Not all but at least Ubuntu, Debian and one of the redhat-based ones?
> 
> I've tested `lsb_release -s` on Pop! (Debian derivative), CentOS-7, Mint, and Ubuntu. I can test it on Debian as well, if you would like.

No, that's fine then.

PR #472, opened 2022-04-18
added -y flag for dnf check-update

mlongmiresqa:
Hello,

I noticed a missing `-y` flag when running `dnf check-update` in the install script. Without this flag you cannot use the auto approve flag like  `./installfog.sh -y` on a fresh OS install due to having to accept the GPG key for the Remi repo.

This flag is present in the similar command `yum -y check-update` so I suspect this was just a miss when adding support for DNF unless there's some difference between the yum and dnf versions of this command I'm not aware of.

### DIFF
--- a/lib/redhat/config.sh
+++ b/lib/redhat/config.sh
@@ -45,7 +45,7 @@ case $linuxReleaseName in
             [[ -z $packageinstaller ]] && packageinstaller="dnf -y install"
             [[ -z $packagelist ]] && packagelist="dnf list"
             [[ -z $packageupdater ]] && packageupdater="dnf -y update"
-            [[ -z $packageUpdate ]] && packmanUpdate="dnf check-update"
+            [[ -z $packageUpdate ]] && packmanUpdate="dnf -y check-update"
             [[ -z $repoenable ]] && repoenable="dnf config-manager --set-enabled"
         else
             [[ -z $packageinstaller ]] && packageinstaller="yum -y install"


mastacontrola:
To my knowledge -y is confirm for install, but check-update didn’t have the requirement as all it does is look for if there are programs to be updated which is why I didn’t bother adding it. If it works it works. Sorry for not adding it and if it was causing issues.

wayneworkman:
I can confirm what @mastacontrola stated. The daily install tests use the -y flag for silent installation and work fine on distributions that use dnf.

Sebastian-Roth:
@mlongmiresqa Can you give us some more context please. Which Linux OS and version do you try this on and which version of FOG did you use? Can't say for sure why the `dnf` call is missing the -y flag. I can see this causing trouble on systems where `dnf update` hasn't been run for a while and repo information (and possibly signing keys) need updating before it can do the update-check.

@mastacontrola @wayneworkman Can you see if causing trouble if we add -y? I don't - though I have to admit that I have not tested or even looked at it closely. Just stating from my guts.

mlongmiresqa:
Hi @Sebastian-Roth -- This is with the latest FOG from the repo and I've run into this on both Rocky Linux and Fedora (I think it was 34, might have been 35). This is not an issue on Centos7/RHEL7 as the corresponding yum command includes the flag. You're correct that the issue is when either `dnf update` hasn't been run in a while **or** just not have run it at all. 

I'm working on automating a FOG server set up with Terraform and Ansible. I run into this issue after copying over my .fogsettings file and running `./foginstall.sh -y` as the keys from the remi repo need to added, but since `dnf check-update` doesn't have the `-y` flag the install can't complete as the system is waiting for a response whether to accept the GPG key or not when the repos are being updated.  

My current work around is just to use ansible to edit the line in this file to add the `-y` flag before running `./foginstall.sh -y` which fixes my issue and the install completes like normal, but it would be nice to not have to edit that.

Sebastian-Roth:
@mlongmiresqa Thanks for the update. Well that commit was quite a while back. Didn't even remember that one.

PR #480, opened 2022-05-18
Update bootmenu.class.php

flagoda:
Add REBOOT options to the Host Bios Exit Type and Host EFI Exit Type.
Use case: Computer boot on PXE only with wakeonlan and are encrypted with Bitlocker.


Sebastian-Roth:
Closing this in favor of #482 to add this to dev-branch as we don't accept PRs on master.

PR #482, opened 2022-05-18
Update bootmenu.class.php

flagoda:
Add REBOOT menu on Host Bios Exit Type and Host EFI Exit Type.
Use case: computer boot on PXE only when wakeonlan and have Bitlocker enable. Because with refind_efi it's ask the recovery key.
REBOOT do when no task boot on the default system (Windows, Linux etc)


Sebastian-Roth:
@flagoda Thanks for the PR. Merging it as is.

PS: May I ask you to combine several changes belonging together in one PR next time?

PR #489, opened 2022-08-15
Fix bug in iPXE embedded scripts which prevents drive boot method EXI…

Piotr86PL:
…T from functioning


Piotr86PL:
Should now work everything. Tested on few computers with different BIOS settings.

PR #494, opened 2022-09-21
Add files via upload

morvany:


morvany:
change version php and change package for php-gettext for php-php-gettext

lib/ubuntu/config.sh
```diff
@@ -1,4 +1,3 @@
-#!/bin/bash
```
Sebastian-Roth, line 1: Why did you remove the bash shebang?

lib/ubuntu/config.sh
```diff
@@ -13,30 +12,32 @@
 #   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 #   GNU General Public License for more details.
 #
+
```
Sebastian-Roth, line 15: Any need to add those empty lines? Just want to keep unnecessary changes at a low.

lib/ubuntu/config.sh
```diff
@@ -13,30 +12,32 @@
...
     sysvrcconf="sysv-rc-conf"
-    phpgettext="php-gettext"
+    phpgettext="php-php-gettext"
     case $OSVersion in
         8)
             php_ver="5"
             ;;
         9)
             php_ver="7.0"
-            x="*php5*"
```
Sebastian-Roth, line 30: Do you understand what this variable is used for? It's used in [line 42](https://github.com/FOGProject/fogproject/blob/dev-branch/lib/ubuntu/config.sh#L42) to remove older versions of PHP that might have been installed earlier. Don't think we want to remove php7 for Debian 9. Same goes for Debian 10.

lib/ubuntu/config.sh
```diff
@@ -13,30 +12,32 @@
...
         10)
             php_ver="7.3"
-            x="*php5* *php7.0*"
+            x="*php7* *php7.0*"
             ;;
         11)
             php_ver="7.4"
-            x="*php5* *php7.0* *php7.3*"
-            phpgettext="php${php_ver}-gettext"
+            x="*php7* *php7.0* *php7.3*"
```
Sebastian-Roth, line 39: Probably removing *php5* is not really necessary because no one ever had PHP5 installed on Debian 11. It's more or less just copy&paste from older versions. But changing that to *php7* does not make much sense to me because we have *php7.0* and *php7.3* in that list as well. If *php7* is in that list a re-run of the FOG installer would purge PHP and reinstall it for no good reason.

lib/ubuntu/config.sh
```diff
@@ -48,11 +49,11 @@ if [[ $linuxReleaseName_lower == +(*bian*) ]]; then
     fi
 elif [[ $linuxReleaseName_lower == +(*ubuntu*|*mint*) ]]; then
     DEBIAN_FRONTEND=noninteractive apt-get purge -yq sysv-rc-conf >/dev/null 2>&1
-    phpgettext="php-gettext"
+    phpgettext="php-php-gettext"
     case $OSVersion in
         22)
             php_ver="8.1"
-            phpgettext="php${php_ver}-gettext"
+            phpgettext="php-php${php_ver}-gettext"
```
Sebastian-Roth, line 56: The package php-php8.1-gettext does not exist in the Ubuntu repo. On the other hand php8.1-gettext does not seem to exist either. I guess this is something from the old days when FOG used the ondrej repos. Should really look this.

lib/ubuntu/config.sh
```diff
@@ -63,7 +64,7 @@ elif [[ $linuxReleaseName_lower == +(*ubuntu*|*mint*) ]]; then
                     php_ver="7.4"
                     ;;
             esac
-            phpgettext="php${php_ver}-gettext"
+            phpgettext="php-php${php_ver}-gettext"
```
Sebastian-Roth, line 67: Same as above.

lib/ubuntu/config.sh
```diff
@@ -76,12 +77,12 @@ elif [[ $linuxReleaseName_lower == +(*ubuntu*|*mint*) ]]; then
             ;;
         16)
             sysvrcconf="sysv-rc-conf"
-            php_ver="7.0"
+            php_ver="7.4"
```
Sebastian-Roth, line 80: I don't think Ubuntu 16.04 defaults to using PHP 7.4. On the other hand we should properly just remove that part as 16.04 only receives updates through ESM (payed) and is not officially supported anymore.

lib/ubuntu/config.sh
```diff
@@ -76,12 +77,12 @@ elif [[ $linuxReleaseName_lower == +(*ubuntu*|*mint*) ]]; then
...
             ;;
         *)
             sysvrcconf="sysv-rc-conf"
-            php_ver="7.1"
-            x="*php5* *php-5*"
+            php_ver="7.4"
```
Sebastian-Roth, line 84: Switching to PHP 7.4 as a default is probably a good idea.

lib/ubuntu/config.sh
```diff
@@ -76,12 +77,12 @@ elif [[ $linuxReleaseName_lower == +(*ubuntu*|*mint*) ]]; then
...
+            x="*php7.4* *php-7.4*"
```
Sebastian-Roth, line 85: Though putting php7.4 on the purge/removal list is not great. See my comment above on the x-variable.

lib/ubuntu/config.sh
```diff
@@ -115,21 +116,21 @@ elif [[ $linuxReleaseName_lower == +(*ubuntu*|*mint*) ]]; then
...
-[[ $php_ver == 5 ]] && php_verAdds="-5.6"
-[[ $php_ver != 5 ]] && phpcmd="php" || phpcmd="php5"
+[[ $php_ver == 5 ]] && php_verAdds="-7.4"
+[[ $php_ver != 5 ]] && phpcmd="php" || phpcmd="php7.4"
 [[ -z $phpfpm ]] && phpfpm="php${php_ver}-fpm"
 [[ -z $phpldap ]] && phpldap="php${php_ver}-ldap"
 [[ -z $phpcmd ]] && phpcmd="php"
-case $linuxReleaseName_lower in
-    *ubuntu*|*bian*|*mint*)
+case $linuxReleaseName in
```
Sebastian-Roth, line 127: You seem to have an older version of the files. Please make sure you use the very latest before sending in a pull request!

lib/ubuntu/config.sh
```diff
@@ -115,21 +116,21 @@ elif [[ $linuxReleaseName_lower == +(*ubuntu*|*mint*) ]]; then
...
+    *[Uu][Bb][Uu][Nn][Tt][Uu]*|*[Bb][Ii][Aa][Nn]*|*[Mm][Ii][Nn][Tt]*)
         if [[ -z $packages ]]; then
             x="mysql-server"
-            eval $packageQuery >>$error_log 2>&1
+            eval $packageQuery >>$workingdir/error_logs/fog_error_${version}.log 2>&1
```
Sebastian-Roth, line 131: Same as above.

lib/ubuntu/config.sh
```diff
@@ -170,6 +171,8 @@ fi
 [[ -z $dhcpconfig ]] && dhcpconfig="/etc/dhcp3/dhcpd.conf"
 [[ -z $dhcpconfigother ]] && dhcpconfigother="/etc/dhcp/dhcpd.conf"
 [[ -z $tftpdirdst ]] && tftpdirdst="/tftpboot"
+[[ -z $tftpconfig ]] && tftpconfig="/etc/xinetd.d/tftp"
```
Sebastian-Roth, line 174: Same as above.

Sebastian-Roth (commented):
A fair amount of questions arise from my point of view. To start of I am wondering if you are using an older version of that file. See the last three comments. As well many of the changes just don't add up for me.

Sebastian-Roth:
@morvany Just in case you are interested, I made some major changes to clean up the version chaos in the installer scripts, see 178fc4c66.

PR #500, opened 2022-10-12
Check if variable is set before calling count (PHP 8 compatibility)

abulhol:
This fixes the following warnings and errors when trying to use the login route programmatically:
```
 PHP Warning:  Undefined variable $http_query in /var/www/fog/lib/pages/processlogin.class.php on line 110
 PHP Fatal error:  Uncaught TypeError:  count():  Argument #1 ($value) must be of type Countable|array, null given in /var/www/fog/lib/pages/processlogin.class.php:
Stack trace:
#0 /var/www/fog/lib/pages/processlogin.class.php(194):
 ProcessLogin->_setRedirMode()\n#1 /var/www/fog/management/index.php(31):
 ProcessLogin->processMainLogin()\n#2 {main}\n  thrown in /var/www/fog/lib/pages/processlogin.class.php on line 110
```

Btw regarding the earlier commit with `array_values`  visible here, you have already merged that before.


Sebastian-Roth:
@abulhol Great, thanks for this PR. I have not tested the API at all yet. Please send in further PRs or issue reports if you find more problems with PHP 8.

abulhol:
For the record, this was not for API use. I have old code that doesn't use the API, but instead logs in via the standard route  and stores a cookie. This wasn't working anymore.

PR #503, opened 2022-10-21
Add Windows 11 to 'os'

flagoda:
Add Windows 11 osid

### DIFF
--- a/packages/web/commons/schema.php
+++ b/packages/web/commons/schema.php
@@ -2738,6 +2738,13 @@ $this->schema[] = array(
     . "VALUES "
     . "('9', 'Windows 10', '')",
 );
+// 181bis
+$this->schema[] = array(
+    "INSERT IGNORE INTO `os` "
+    . "(`osID`, `osName`, `osDescription`) "
+    . "VALUES "
+    . "('11', 'Windows 11', '')",
+);
 // 182
 $this->schema[] = array(
     "UPDATE `pxeMenu` SET `pxeParams`='login\n"


Sebastian-Roth:
@flagoda Thanks for your pull request on adding Windows 11. Though there is a lot more to consider when doing this.

1. This would only work for new installs of FOG but it won't add the Windows 11 definition if upgrading from an older version of FOG. The schema updater reads schema version numbers from `/var/www/fog/lib/fog/system.class.php` and the database and will only add schema changes if system.class.php has a higher version number. As well the schema updater would only add the last schema changes (as many as version numbers differ) - see [databasemanager.class.php line 78 to 125](https://github.com/FOGProject/fogproject/blob/master/packages/web/lib/db/databasemanager.class.php#L87) and [schemaupdaterpage.class.php line 173ff](https://github.com/FOGProject/fogproject/blob/master/packages/web/lib/pages/schemaupdaterpage.class.php#L173).
2. The `osid` is mostly used by the FOS inits to decide weather certain steps need to be done one way or the other (e.g. Win XP vs. Win 10 vs. Linux vs. macOS). So if we add a new `osid` for Win 11 a fair amount of changes need to be made to the FOS scripts as well - [1](https://github.com/FOGProject/fos/blob/master/Buildroot/board/FOG/FOS/rootfs_overlay/usr/share/fog/lib/funcs.sh#L80), [2](https://github.com/FOGProject/fos/blob/master/Buildroot/board/FOG/FOS/rootfs_overlay/usr/share/fog/lib/funcs.sh#L1265), [3](https://github.com/FOGProject/fos/blob/master/Buildroot/board/FOG/FOS/rootfs_overlay/usr/share/fog/lib/funcs.sh#L1516), [4](https://github.com/FOGProject/fos/blob/master/Buildroot/board/FOG/FOS/rootfs_overlay/usr/share/fog/lib/funcs.sh#L1642), [5](https://github.com/FOGProject/fos/blob/master/Buildroot/board/FOG/FOS/rootfs_overlay/usr/share/fog/lib/funcs.sh#L2285) just to name a few.
3. Currently we a pretty cautious with adding changes to the schema because two versions of FOG are in development (dev-branch) and (working-1.6). The schema diverged at some stage and it needs thorough consideration when changing the schema.

In general I am wondering if adding Windows 11 as it's own `osid` is actually a good idea. As long as the FOS inits don't need to handle Win 10 and Win 11 differently I would rather not add a new `osid`. Maybe we should rather think about changing the name in the drop down list from "Windows 10" to "Windows 10/11".

flagoda:
To deploy Win 11, i need have an osid different than Win 10, because for example sysprep/setupcomplete are different between 10 and 11 version.
I will cheat in my own scripts, I thought it would have been easier like that.

Issue #507, opened 2022-10-23
Stripped kernel version information in FOG reporting


Sebastian-Roth:
Some systems seem to report shorter kernel version strings than others, e.g.
```
5.15.68 (buildkite-agent@Tollana) #1 SMP Sun Oct 9 06:54:22 CDT 2022
```
vs.
```
5.15.68 (buildkite-agent@Tollana) #1 SMP Sun Oct 9 06:54:22 CD
```

Although those are the exact same kernel binaries those are reported as being two different ones.

Sebastian-Roth:
Taking a closer look at the database data I found that CentOS 7 as well as Ubuntu 16.04 systems return a stripped version string.

These are the options we have:
* Use different command to get the version information like `strings filename | grep ...` ([1](https://stackoverflow.com/a/36889965))
* Compile our own tool to get the version ([2](https://stackoverflow.com/a/11179559)) - FOG installs gcc compiler to build udpcast anyway.
* Strip down the version info even more - only use `5.15.68` or maybe `5.15.68 (buildkite-agent@Tollana)`.

At first I thought we would loose information if we toss out the exact build date. But the more I think about it I find it's probably a good idea to do. Why should we distinguish between same kernel versions build on different dates?

I know we discussed what to use from the kernel version when we started this. But seeing the results now I tend to only use less information.

wayneworkman:
I think the simplest would be the 3rd option, to strip down the information even more. Every single one is returning a `#` immediately after the information we want, so that can be a delimiter.

Sebastian-Roth:
@wayneworkman's PRs merged, thanks.

Sebastian-Roth:
@wayneworkman I am now wondering if we should even leave out the prepended filename of the binary. Essentially we want the kernel version and it doesn't matter if it's a 32 or 64 bit kernel I think. What's your opinion on this?

wayneworkman:
Originally we included the filenames so we could distinguish 32 from 64. We can take out the filename too, if that is your preference. I'll submit PRs.

Sebastian-Roth:
@wayneworkman I have thought about it for a bit more and still can't see why we should need to know if it's a 32 vs. 64 bit kernel. We always release both at the same time and so there should never be a difference in how those are configured.

Thanks for the new PRs!

PR #513, opened 2022-11-05
Do not prepend filename to kernel information

wayneworkman:


wayneworkman:
This further helps with issue #507

PR #520, opened 2022-11-29
Update multicasttask.class.php

Magelfik:
Allows all partitions to be set at an equal timeout from udp-sender --max-wait option

Full explanation here : https://github.com/FOGProject/fogproject/issues/521


Sebastian-Roth:
@Magelfik Great, thanks!

PR #530, opened 2023-01-04
Do not install sysv-rc-conf on Debian hosts

### LINKED ISSUE #531: Fix dev-branch installer to work on Debian Testing ("Bookworm")
The FOG installer in dev-branch can install FOG server on Debian Testing ("Bookworm"), when invoked as follows:

    linuxReleaseName=Debian OSVersion=12 OSMinorVersion=0 ./installfog.sh

Unfortunately, the installer installs the `sysv-rc-conf` package, which breaks systemd. This prevents the FOG service units from working, effectively crippling the FOG server.

I have created a pull request (#530) that proposes a simple solution: Modify the installer to not install `sysv-rc-conf` on Debian hosts. This works because `sysv-rc-conf` is not used on Debian hosts anyway, since service control is done using systemd tooling.

stevemoca:
* Debian hosts use systemd to enable services, so `sysv-rc-conf` is not needed.
* systemd on Debian Testing ("Bookworm") breaks when `sysv-rc-conf` is installed; this PR prevents that.
* Closes #531

### DIFF
--- a/lib/ubuntu/config.sh
+++ b/lib/ubuntu/config.sh
@@ -17,7 +17,9 @@
 #   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 [[ -z $packageQuery ]] && packageQuery="dpkg -l \$x | grep '^ii'"
-if [[ $linuxReleaseName_lower == +(*bian*) ]]; then
+if [[ $linuxReleaseName_lower == +(*debian*) ]]; then
+    : # Debian hosts use systemd, so sysv-rc-conf is not needed
+elif [[ $linuxReleaseName_lower == +(*bian*) ]]; then
     sysvrcconf="sysv-rc-conf"
 elif [[ $linuxReleaseName_lower == +(*ubuntu*|*mint*) ]]; then
     DEBIAN_FRONTEND=noninteractive apt-get purge -yq sysv-rc-conf >/dev/null 2>&1
--- a/lib/ubuntu/config.sh
+++ b/lib/ubuntu/config.sh
@@ -17,25 +17,13 @@
 #   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 [[ -z $packageQuery ]] && packageQuery="dpkg -l \$x | grep '^ii'"
-if [[ $linuxReleaseName_lower == +(*debian*) ]]; then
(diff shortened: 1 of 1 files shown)


lib/ubuntu/config.sh
```diff
@@ -17,7 +17,9 @@
 #   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 [[ -z $packageQuery ]] && packageQuery="dpkg -l \$x | grep '^ii'"
-if [[ $linuxReleaseName_lower == +(*bian*) ]]; then
+if [[ $linuxReleaseName_lower == +(*debian*) ]]; then
+    : # Debian hosts use systemd, so sysv-rc-conf is not needed
+elif [[ $linuxReleaseName_lower == +(*bian*) ]]; then
```
Sebastian-Roth, line 22: I don't think any of the other `*bian*` distros is using SystemV anymore either. As well Ubuntu 16.04 does not actually need it I reckon and it's EOL anyway. So let's remove that whole part - line 20 to 29!

lib/ubuntu/config.sh
```diff
@@ -74,4 +76,4 @@ fi
 [[ -z $ftpconfig ]] && ftpconfig="/etc/vsftpd.conf"
 [[ -z $snapindir ]] && snapindir="/opt/fog/snapins"
 [[ -z $dhcpd ]] && dhcpd="isc-dhcp-server"
-[[ -z $dhcpname ]] && dhcpname="isc-dhcp-server"
\ No newline at end of file
+[[ -z $dhcpname ]] && dhcpname="isc-dhcp-server"
```
Sebastian-Roth, line 79: Do you mind leaving the new line at the end of the file in place?

Sebastian-Roth (changes requested):
Let's get this one done first and then see if we remove all references to SystemV from the installer later on.

lib/ubuntu/config.sh
Sebastian-Roth: Absolutely!

lib/ubuntu/config.sh
Sebastian-Roth: Whoops, I had this upside down in my head. Thanks.

PR #539, opened 2023-01-29
Add authentication check in getfiles to prevent file enumeration

0x41c:
My branch merged very weirdly, to prevent commit clutter merge as squashed commit :)


Sebastian-Roth:
@0x41c Can't find anywhere in the code that might be upset by this check. Though I don't know all of the code and can't be absolutely sure.

PR #540, opened 2023-01-29
Introduce log sanitization to `logtoview.php` so logs are not parsed as html

0x41c:


packages/web/status/logtoview.php
```diff
@@ -25,20 +25,6 @@
...
-    exit;
-}
-if (!(isset($_POST['file'])
-    && is_string($_POST['file']))
-) {
-    echo json_encode(_('Invalid File'));
-    exit;
-}
-$file = '';
-$lines = '';
```
0x41c, line 41: These were removed because of duplication

0x41c:
I've also added fixes for the possibility of $currentUser being null when checking for authentication status on certain endpoints. I meant to make a new branch for those changes so that they could be merged separately, but the quicker it gets fixed, the less time there is for people to have these issues show up in prod.

0x41c:
So after finding some errors show up, I did a little digging. How come we're unauthenticated at this point?
https://github.com/FOGProject/fogproject/blob/3ed1818465eb59f12b0602e1605218847935e07a/packages/web/lib/fog/storagenode.class.php#L225-L236

The reason I ask this is because it seems like the getfiles endpoint is returning nothing here.

packages/web/service/getversion.php
```diff
@@ -43,11 +43,15 @@
     );
 } elseif (isset($_REQUEST['url'])) {
 
+    if (is_null($currentUser))
```
mastacontrola, line 46: Is there a reason we're using goto? Normally this isn't best practice.

Also,

If (is_null($currentUser)) shouldn't be necessary as currentUser should always be set. Even if it's not, I don't think we're exposing anything to a person by showing the version of the FOG server or Client. Mind you the idea of GetVersion in our context here is to allow the user to know what version exists on the server vs what exists on the device in question (who wouldn't necessarily have a login mechanism to the server)

packages/web/status/kernelvers.php
```diff
@@ -42,7 +46,7 @@
         echo $response;
         unset($response);
     }
-    
+
```
mastacontrola, line 49: Similar here,

What exposure is there showing a person who's unauthenticated seeing the kernel versions?

packages/web/status/logtoview.php
mastacontrola: Where are these duplicated?

packages/web/service/getversion.php
0x41c: Actually in this scope where the URL parameter is defined, they should be authenticated since it causes the server to perform a request to the URL. If however you're saying that we should give the version regardless of the login-status and just not perform the request when the `url` parameter is defined, I can only see it as being a little misleading. Regardless, it's doable.

packages/web/status/logtoview.php
0x41c: The conditions that check the post parameters `ip` and `file` were evaluated twice, once after initial evaluation. Therefore they aren't needed at that point. Both `$file` and `$lines` are treated as function parameters for `val` and are not used as globally scoped variables within the function. Therefore defining them up at the top of the file has no effect.

`$file` and `$lines` can be initialized after their respective conditions lower down in the file since they aren't used before then anyways.

packages/web/status/kernelvers.php
0x41c: I believe both here and in getversion.php you're missing the scope of the check. This isn't preventing a normal unauthenticated user from viewing the kernel version. It prevents them from making the server request a URL when they aren't able to authenticate themselves. By omitting `url` in their POST data, they can still view the version without a hitch.

PR #552, opened 2023-03-05
Fix location based snapin download by adding location URL

### LINKED ISSUE #371: Repair snapin download not working on HTTPS enabled servers
The code added in commit d0162c82 breaks snapin download on HTTPS enabled servers and is therefore not appropriate in all cases. There is no quick fix for that as we don't have the information about a node being setup as HTTP or HTTPS in the database object for each storage node. This was missed when adding HTTPS to FOG and needs some thorough consideration.

The quick fix is to just set "https" in the code or remove those five lines that were added when trying to quickly fix the issue with snapins not downloading in locations.

Sebastian-Roth:
based on protocol information now stored in the database for each location storage node individually.
Fixes #371

### DIFF
--- a/Release Notes.MD
+++ b/Release Notes.MD
@@ -81,6 +81,7 @@ To view the full list of commits with the code changes introducing these fixes a
 
 ### Bug Fixes
 
+* Fix location based snapin download (#371)
 * Fix group cron style task scheduling (#544)
 * Fix incorrect array index name in location plugin code.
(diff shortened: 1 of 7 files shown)


Sebastian-Roth:
Tests look fine on existing installs ([manual DB adjustments needed](https://github.com/FOGProject/fogproject/issues/371#issuecomment-1455136385)) as well as fresh installs.

PR #570, opened 2023-07-28
FIX #567 ldap plugin with PHP 8

### LINKED ISSUE #567: LDAP Plugin - PHP 8 compatibility
LDAP Plugin currently not working with PHP 8.

Upgrade PHP to PHP 8.x

LDAP just provide authentication via AD.

 - FOG version: latest dev code
 - OS: Debian 12

tomamplius:
see #567

### DIFF
--- a/packages/web/lib/plugins/ldap/class/ldap.class.php
+++ b/packages/web/lib/plugins/ldap/class/ldap.class.php
@@ -232,7 +232,10 @@ class LDAP extends FOGController
         /**
          * Ensure any trailing bindings are removed
          */
-        @$this->unbind();
+        try {
+            $this->unbind();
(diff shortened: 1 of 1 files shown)


Sebastian-Roth:
@tomamplius Thanks for sending in the PR. I am wondering if there is a nicer way to just check if "this" is a valid LDAP connection object instead of catching any TypeError exception. But on the other hand it seems like it has been like this for a long time using the "@" notation and so I am fine with this fix.

But I need to change the base branch to "dev-branch" as this is the one we collect all such changes in before merging into master.

Sebastian-Roth (commented):
Looking fine from my view although it just masks the issue.

PR #634, opened 2024-08-30
Fix inventory panels

rluzuriaga:


mastacontrola:
Closed as this is handled in a different way, but quite at the same time similarly.

PR #635, opened 2024-09-02
updated ldap plugin

geotsot:
- added bind ldap on search without group matching
- concatenate groups to match the provided Group Member Attribute instead of hardcoded name=
- small correction on dropdown menu for Subtree and below


geotsot:
@mastacontrola sorry for the spam commits... tried to resolve the conflicts but didn't succeed.

darksidemilk:
I would suggest you stash/save your changes to the ldap stuff and then do a fresh fork and try again. 
I would make sure to fork all branches  and run this too
```
git fetch --all
git config core.hooksPath .githooks
```
Which will implement the githook in .githooks that auto increments the version based on the number of commits since the last master version was released.

darksidemilk:
You could also try to just sync your current fork. Something along those lines should resolve the conflict.

PR #642, opened 2024-09-08
ldap plugin update

geotsot:
ldap plugin after resync of my dev-branch (ref: https://github.com/FOGProject/fogproject/pull/635)


mastacontrola:
I have pushed what I suspect will fix the issues you are having with the PR here as well as address the concerns you presented with only the main difference being the forced bindDN checks you have on 521-571. I have left that piece out. If it is required please first think about what is happening there. The reason we don't FORCE a bindDN user/password is because not all environments require it. Your current setup is effectively enforcing it to everybody.

With all the issues you ran into with merge conflicts and versioning I have made the changes in dev-branch and pushed them. I hope you know I do appreciate the effort you've taken and given, I hope, proper respect to your code and contributions here. I am closing this PR as I think it is now covered and without all the hassle of the manually adjusted FOG_VERSION and all that.

Please test and let me know if there's any issues.

PR #654, opened 2024-11-12
fix missing cp introduced in cd89b56f5e

osiktech:
in commit cd89b56f5e a bug was introduced which broke copying of arm64-efi files.


darksidemilk:
Sorry that we didn't see this before the release just now.

PR #698, opened 2025-03-13
wrong variables used in addModule e RemoveModule - API call to edit host

sephcron:
Variables was used in wrong methods:

$modulesToRem was used in addModule
$modulesToAdd was uses in removeModule


sephcron:
additionally, this problem occurs when trying to edit modules through the API call PUT /fog/host/{id}/edit with body  like {"modules": [$x,$y,$z]}

darksidemilk:
Nice catch. I changed your base branch for merge to dev-branch as we don't commit changes to stable manually, only via the github action/workflow for releases.

PR #699, opened 2025-03-13
wrong method used to add modules - API call to edit group

sephcron:
AddPrinter method was used instead of AddModule


sephcron:
additionally, this problem occurs when trying to edit modules through the API call PUT /fog/group/{id}/edit with body  like {"modules": [$x,$y,$z]}

darksidemilk:
Looks good here too. Changed the base branch to dev-branch. The automated release to stable will occur on the 15th so you'll see these changes show up on the 15th. I'll take care of incrementing the version in a separate commit for the 2 PRs. Thank you for your contribution!

PR #702, opened 2025-03-14
remove id as required field in site plugin

sephcron:
API calls to create "site" and "sitehostassociation" always return error: "Required database field is empty: id":

POST /fog/site {"name": "site name"}
POST /fog/sitehostassociation {"siteID": 1, "hostID": 2}

I don't know if there is a reason for this field to be required, but by making this change the API calls work.

Taking the location plugin as a reference, which has a similar functionality, the id field is not required.

### DIFF
--- a/.githooks/pre-merge-commit
+++ b/.githooks/pre-merge-commit
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/sh
 #
 # An example hook script to verify what is about to be committed.
 # Called by "git merge" with no arguments.  The hook should
--- a/packages/web/lib/fog/system.class.php
+++ b/packages/web/lib/fog/system.class.php
@@ -53,7 +53,7 @@ class System
     public function __construct()
     {
         self::_versionCompare();
-        define('FOG_VERSION', '1.5.10.1647');
+        define('FOG_VERSION', '1.5.10.1648');
         define('FOG_SCHEMA', 273);
         define('FOG_BCACHE_VER', 141);
         define('FOG_CLIENT_VERSION', '0.13.0');
--- a/packages/web/lib/fog/system.class.php
+++ b/packages/web/lib/fog/system.class.php
@@ -53,7 +53,7 @@ class System
     public function __construct()
     {
         self::_versionCompare();
-        define('FOG_VERSION', '1.5.10.1646');
+        define('FOG_VERSION', '1.5.10.1647');
         define('FOG_SCHEMA', 273);
         define('FOG_BCACHE_VER', 141);
         define('FOG_CLIENT_VERSION', '0.13.0');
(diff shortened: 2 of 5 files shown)


sephcron:
added suport to define siteID on calls to create or edit hosts and groups:

POST /fog/host
{"name": "host_name", "macs": ["aa:bb:cc:dd:ee:ff"], "siteID": 1}

PUT /fog/host/1
{"siteID": 1}

POST /fog/group
{"name": "group name", "hosts": [1, 2, 3], "siteID": 1}

PUT /fog/group/1
{"siteID": 1}

GET /fog/host
Result json: { {... , "siteID": 1}, {... , "siteID": 2} }

GET /fog/host/1
Result json: { ... , "siteID": 1}

darksidemilk:
@sephcron If you don't mind me asking, how are you accessing the api? Just curious if you're using the [FogApi Powershell Module](https://github.com/darksidemilk/FogApi) or if you're building your own tool.

Also, for incrementing the version run `git config core.hooksPath .githooks` to enable this hook -> https://github.com/FOGProject/fogproject/blob/dev-branch/.githooks/pre-commit. 
You may need to run `git fetch --all` for it to work properly, sometimes I find I have to checkout and do a pull on the master branch and stable branches once on a fresh clone then go back to the dev-branch. The pre-commit script does a count of commits since the last master branch baseline (1.5.0) to the dev-branch to get the revision we track. You don't have to do that, we can always fix it in another commit, just wanted to share the tip.

I tested creating a site on fog 1.6 and didn't get the ID error. So that aspect is already fixed there, so no problem fixing it on the dev-branch too.

I like the hook add with the site plugin, but have you tested what happens when the site plugin is removed/uninstalled? @mastacontrola might know better if that's just something handled elsewhere in the plugin system. We'll probably want to add this functionality to 1.6 as well.

Thank you for your contribution!

sephcron:
This is the first time I've made contributions, I'm still learning how to deal with git branch and pull requests, so any tips are welcome.

I'm consuming the FOG API with PHP, to communicate with another web application.

I tested removing the plugin and everything seems to be OK, the field is not returned on GET and ignored on POST/PUT.

I have plans to add the same functionality to the location plugin API if that's okay.

Just to confirm, there is no need for me to increase the version on commit?

English is not my native language, so sorry for the text.

darksidemilk:
Adding the same functionality to the location plugin would be great, I'd do a separate PR for that. 
If you want to try adding the same hook to the working-1.6 branch, that would be good too. 

As a side note, I would also recommend trying out installing fog from the 1.6 branch as there are some API changes and improvements. For example all return objects change to a unified `data` property instead of a named property. i.e. instead of a return value of `hosts` on `/fog/host` the resulting hosts are in `data`. Wouldn't want your site integrations to break when we make that the new stable version which should be soon-ish.  

You only need to adjust the version if you want to try using the githook, we'll redo it otherwise.

Also, your English is great. 

I'll pull this in shortly.

darksidemilk:
Dang it, I messed it up. I was trying to bring in the merge of the PR via command line so my githook would increment the version. It did work but it also didn't because the PR didn't auto close and it says I wrote it. I'll fix it.

PR #703, opened 2025-03-14
added suport to set locationID on host and group API calls

sephcron:
added suport to set locationID on API calls to create or edit hosts and groups, and returning it on get:

POST /fog/host
{"name": "host_name", "macs": ["aa:bb:cc:dd:ee:ff"], "locationID": 1}

PUT /fog/host/1
{"locationID": 1}

POST /fog/group
{"name": "group name", "hosts": [1, 2, 3], "locationID": 1}

PUT /fog/group/1
{"locationID": 1}

GET /fog/host
Result json: { {... , "locationID": 1}, {... , "locationID": 2} }

GET /fog/host/1
Result json: { ... , "locationID": 1}

It is important to note that although it is possible to associate a host with multiple locations and sites through the sitehostassociation and locationassociation direct API calls, the user interface behavior when updating the host is to allow only 1 site and 1 location, and the others are removed. I am imitating this behavior in these changes, keeping only 1 location and 1 site for each host (site was done in previous PR).

Also, when testing I found a strange behavior when trying to create a location, both through the API and the UI. In some cases the plugin will update an existing record instead of creating it. This happens because of the INSERT INTO ... ON DUPLICATE KEY UPDATE clause having other UNIQUE INDEXES in the table, and missing constraints validations on save calls. I don't know if this is actually a problem, but it may seem strange to users.

### DIFF
--- a/packages/web/lib/fog/system.class.php
+++ b/packages/web/lib/fog/system.class.php
@@ -53,7 +53,7 @@ class System
     public function __construct()
     {
         self::_versionCompare();
-        define('FOG_VERSION', '1.5.10.1650');
+        define('FOG_VERSION', '1.5.10.1653');
         define('FOG_SCHEMA', 273);
         define('FOG_BCACHE_VER', 141);
         define('FOG_CLIENT_VERSION', '0.13.0');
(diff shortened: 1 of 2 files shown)


darksidemilk:
Merged in this commit:
https://github.com/FOGProject/fogproject/commit/ae883b6a62f614bd0b4a492439d66aa325b91959

PR #706, opened 2025-03-19
fix object operator (site plugin)

sephcron:
I found this error on site plugin, it prevents a host from being completely deleted. 

Although the host is no longer displayed in the user interface (probably because the associated MAC addresses have been deleted), the orphaned record still exists in the host table, which prevents the hostname from being reused on another host. 

This only happens when the site plugin is installed.


sephcron:
Also, this error occurs when trying to delete the record on the host edit page.

Deleting a host on the host listing page does not cause this error, although deleting a host this way does not seem to remove associations in the siteHostAssoc table (maybe hooks are not triggered this way?).

sephcron:
the error also occurs when trying to delete a host via the API, returning error 500

PR #733, opened 2025-08-20
Add libcurl4t64 support for Debian 13 (Trixie)

delumv:
Debian 13 (Trixie) renamed the libcurl4 package to libcurl4t64, but the FOG installer still checks for libcurl4, causing installation to fail.

This change adds a version check for OS >= 13 and replaces libcurl4 with libcurl4t64 in the package list, similar to how it's already handled for Ubuntu 22.04+.

Fixes installation on Debian 13.


mastacontrola (approved):
Thanks for the help.

PR #739, opened 2025-09-01
Update functions.sh - Fix debian 13

Mandrilux:


darksidemilk:
Looks similar to https://github.com/FOGProject/fogproject/pull/733

Need to take a closer look.

darksidemilk:
Looks like you branched from stable rather than dev. We put all changes in dev-branch and it gets released into stable one a month on the 15th (provided all tests pass). It looks like this change is the same as the pull request we just brought in last week. Let me know if I'm missing something that's different @Mandrilux

darksidemilk:
Looks like you put libcurl3 instead of libcurl4?

mastacontrola:
I'm closing this pull request.

We already have a fix in dev-branch that will address this once it moves to stable, as well it's been pushed into working-1.6. This specific pull request would actually fail because:
The if stanza right before it would change libcurl3 -> libcurl4, so this if stanza addon would fail to actually _do_ anything.

PR #748, opened 2025-10-06
hostgetkey store token before save in db and return

kbats183:
I encountered an issue when trying to pull tasks using `fog/service/hostinfo.php`.

In some cases, `fog/status/hostgetkey.php` generates a new token, saves it to the database, but returns an empty string as the token value.
I believe this happens due to a race condition between storing the token in the database and retrieving it for returning.

This change generates the new token before updating the database and reuses it for both the DB update and the return value.
This ensures the function always returns the correct token.

### DIFF
--- a/packages/web/status/hostgetkey.php
+++ b/packages/web/status/hostgetkey.php
@@ -36,15 +36,16 @@ try {
         throw new Exception(_('Host token is currently in use'));
     }
     if (!FOGCore::$Host->get('token')) {
+        $newToken = FOGCore::createSecToken();
         FOGCore::getClass('HostManager')->update(
             ['id' => FOGCore::$Host->get('id')],
             '',
             [
-                'token' => FOGCore::createSecToken(),
+                'token' => $newToken,
                 'tokenlock' => true
             ]
         );
-        throw new Exception(FOGCore::$Host->get('token'));
+        throw new Exception($newToken);
     }
     if (FOGCore::$Host->isValid() && !FOGCore::$Host->get('tokenlock')) {
         FOGCore::getClass('HostManager')->update(


darksidemilk:
What is your use case for pulling tasks via the php file vs using the api?

________________________________
JJ Fullmer
Senior IT Systems Engineer
[Arrowhead Dental Laboratory]
Email   ***@***.******@***.***>
Phone   (801) 572-7253<tel:+1(801)%20572-7253>
Web     arrowheaddental.com<https://www.arrowheaddental.com/>
________________________________
[Facebook] <https://www.facebook.com/ArrowheadDentalLab>        [Instagram]  <https://www.instagram.com/arrowheaddentallab/>    [YouTube]
CONFIDENTIALITY NOTICE:
This e-mail, including any attachments, may contain confidential information which is intended only for the use of the individual(s) or entities named.
If you receive this e-mail message in error, please immediately notify the sender by e-mail and delete it.
Dissemination, forwarding, printing or copying of this e-mail without prior consent of the sender is strictly prohibited.

________________________________
From: Konstantin Bats ***@***.***>
Sent: Monday, October 6, 2025 3:37:57 PM
To: FOGProject/fogproject ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [External][FOGProject/fogproject] hostgetkey store token before save in db and return (PR #748)


I encountered an issue when trying to pull tasks using fog/service/hostinfo.php.

In some cases, fog/status/hostgetkey.php generates a new token, saves it to the database, but returns an empty string as the token value.
I believe this happens due to a race condition between storing the token in the database and retrieving it for returning.

This change generates the new token before updating the database and reuses it for both the DB update and the return value.
This ensures the function always returns the correct token.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/FOGProject/fogproject/pull/748

Commit Summary

  *   06477ff<https://github.com/FOGProject/fogproject/pull/748/commits/06477ffdc6965d74d68bbd0d08f86460283b8819> hostgetkey store token before save in db and return

File Changes

(1 file<https://github.com/FOGProject/fogproject/pull/748/files>)

  *   M packages/web/status/hostgetkey.php<https://github.com/FOGProject/fogproject/pull/748/files#diff-21c4a429533edde44e3af0159f8e7bfc0b653e13bcdbddffc880429ac8be712c> (5)

Patch Links:

  *   https://github.com/FOGProject/fogproject/pull/748.patch
  *   https://github.com/FOGProject/fogproject/pull/748.diff

kbats183:
I’m pulling tasks via the PHP file using the same approach that FOG uses to [get kernel argument properties during USB boot](https://github.com/FOGProject/fos/blob/master/Buildroot/board/FOG/FOS/rootfs_overlay/bin/fog#L3-L12).

My use case is that I have more than 100 laptops that I want to image via FOG multicast only once (without continuous management through FOG afterward).
I’m using the Quick Registration option to register the hosts, assign an image, and add them to a group — and I’d like to run this whole process without requiring an extra reboot between quick registration and imaging.
I’ve also [patched the kernel](https://github.com/kbats183/fos/blob/tasks-pulling/Buildroot/board/FOG/FOS/rootfs_overlay/bin/fog.pull.task) slightly to support this workflow.

mastacontrola:
I added this as it make sense and is a very likely case. Thank you!

PR #759, opened 2025-11-25
Minor grammar fix for multi-select delete.

ggppjj:
This change resolves my only long-term frustration with Fog, the fact that the grammar on the multi-select delete page's title uses a possessive apostrophe where it shouldn't.

### DIFF
--- a/packages/web/lib/fog/fogpage.class.php
+++ b/packages/web/lib/fog/fogpage.class.php
@@ -1907,7 +1907,7 @@ abstract class FOGPage extends FOGBase
         global $sub;
         global $node;
         $this->title = sprintf(
-            "%s's to remove",
+            "%ss to remove",
             (
                 $this->childClass !== 'Storage' ?
                 $this->childClass :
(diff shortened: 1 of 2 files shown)


darksidemilk:
Thank you! Good catch. 
I don't believe that typo exists in fog 1.6 if you wanted to give that version a try.

ggppjj:
I did a quick pass over 1.6, looks good! I tend to stick with stable for my use-case, so one of these days I'm sure I'll find myself on it. Thanks for the lightning-fast merge!

darksidemilk:
I consider 1.6 stable, we just haven't had time to make the move and need more input from more use cases to be extra sure it's stable. I've been using it in production for a year or 2 (time is hard but it's been quite a while without any issues).

ggppjj:
Gotcha. I use it in VM at my ~80 or so deployed on-site mostly retail point-of-sale end-customer systems to provide easy emergency PXE re-imaging of lanes, so I only ever tend to update things when the IP scheme needs adjustment or similar. Most of the time for me the Fog VM is off and tucked away for when it's needed, which explains my slow uptake on 1.6. I'll check it out some time!

PR #760, opened 2025-12-05
Added ntfy integration

TonyKaito:
Created a new plugin that sends event notifications to an ntfy push notification server, whether ntfy.sh main server or a self-hosted one

### DIFF
--- a/packages/web/lib/fog/system.class.php
+++ b/packages/web/lib/fog/system.class.php
@@ -53,7 +53,7 @@ class System
     public function __construct()
     {
         self::_versionCompare();
-        define('FOG_VERSION', '1.5.10.1734');
+        define('FOG_VERSION', '1.5.10.1735');
         define('FOG_SCHEMA', 273);
         define('FOG_BCACHE_VER', 141);
         define('FOG_CLIENT_VERSION', '0.13.0');
--- a/packages/web/lib/plugins/ntfy/class/ntfy.class.php
+++ b/packages/web/lib/plugins/ntfy/class/ntfy.class.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * The ntfy database and object definer
+ *
+ * PHP version 5
+ *
+ * @category Ntfy
+ * @package  FOGProject
+ * @author   Tony Lam <tonylam5349@gmail.com>
+ * @license  http://opensource.org/licenses/gpl-3.0 GPLv3
+ * @link     https://fogproject.org
+ */
+/**
+ * The ntfy database and object definer
+ *
+ * @category Ntfy
+ * @package  FOGProject
+ * @author   Tony Lam <tonylam5349@gmail.com>
+ * @license  http://opensource.org/licenses/gpl-3.0 GPLv3
+ * @link     https://fogproject.org
+ */
+class Ntfy extends FOGController
+{
+    /**
+     * The ntfy table.
+     *
+     * @var string
+     */
+    protected $databaseTable = 'ntfy';
+    /**
+     * The database fields and commonized items.
+     *
+     * @var array
+     */
+    protected $databaseFields = array(
+        'id' => 'nID',
+        'serverURL' => 'nServerURL',
+        'topicEndpoint' => 'nTopicEndpoint',
+        'credentials' => 'nCredentials',
+    );
+    /**
+     * The required fields
+     *
+     * @var array
+     */
+    protected $databaseFieldsRequired = array(
+        'serverURL',
+        'topicEndpoint',
+    );
+}
--- a/packages/web/lib/plugins/ntfy/class/ntfyexception.class.php
+++ b/packages/web/lib/plugins/ntfy/class/ntfyexception.class.php
@@ -1,0 +1,24 @@
+<?php
+/**
+ * Exception class for ntfy
+ *
+ * PHP Version 5
+ *
+ * @category NtfyException
+ * @package  FOGProject
+ * @author   Tony Lam <tonylam5349@gmail.com>
+ * @license  http://opensource.org/licenses/gpl-3.0 GPLv3
+ * @link     https://fogproject.org
+ */
+/**
+ * Exception class for ntfy
+ *
+ * @category NtfyException
+ * @package  FOGProject
+ * @author   Tony Lam <tonylam5349@gmail.com>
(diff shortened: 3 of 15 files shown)


Neustradamus:
@rluzuriaga, @mastacontrola: Any progress on it?

mastacontrola:
Thanks for this, @TonyKaito! 🙏 The feature is a good idea, and the reason it hasn't merged is that `dev-branch` (1.5.x) isn't really where notification plugins are heading — `working-1.6` is, and it recently gained a **per-plugin schema-migration system** that any new table-owning plugin now has to participate in. So rather than let this stall, I've ported and reworked it onto `working-1.6`.

Branch: [`feature-760-ntfy`](https://github.com/FOGProject/fogproject/tree/feature-760-ntfy) · commit 8fb4099f5

### Context I found while reviewing
The plugin was forked from the older `pushbullet` plugin (the `pushNote`/`_parseRecipient`/`device_iden`/`channel_tag` bits are Pushbullet's API), so it inherited some Pushbullet-isms and predates everything `working-1.6` changed about plugins.

### Flaws fixed
- **Wire protocol** — the handler still spoke Pushbullet's JSON (`type=note&title=…&body=…` form fields). ntfy takes the **message as the POST body** and the **title via a `Title` header**, so as written it published a literal `type=note…` string with no title. Reworked to speak ntfy. (Verified against a live `ntfy.sh` topic: title + message land correctly.)
- **`NftyException` typo**, **bareword array keys** (`$data[email]` → fatal on PHP 8), and a referenced-but-undefined `$this->_apiKey` path — all removed/fixed.
- **`->find()`** doesn't exist on `FOGManagerController` in 1.6 (dedup would've fataled); switched to `distinct()`.
- **Page never registered** — the page loader only discovers `*.page.php` whose class matches the filename. The file was `ntfymanagementpage.class.php` / `NtfyManagementPage`, so `node=ntfy` would 404. Renamed to `ntfymanagement.page.php` / `NtfyManagement`.
- **Unescaped output** — values were echoed straight into `value="…"` attributes; rebuilt on `makeInput`/`makeLabel`/`makeButton`/`formFields`/`makeFormTag` (which run through `Initiator::e()`), added `checkAuthAndCSRF()` + `HTTPResponseCodes` to `addPost`.
- Removed Slack leftovers (`unset($Slack)`, commented blocks) and the manual `FOG_VERSION` bump (the pre-commit hook owns versioning on `working-1.6`).

### Brought up to working-1.6 conventions
- **Schema migrations**: `createSql()` + append-only `schema()` + non-destructive `install()` via `Schema::applyUpdates()` — no more drop-and-recreate on re-install; plugs into the dashboard "update available" flow. See `docs/PLUGIN_SCHEMA_MIGRATIONS.md`.
- Added the missing **API** and **JS** hooks (the JS hook is why `fog.ntfy.*.js` now actually loads); menu hook converted to the `hook_main` pattern; JS split into `fog.ntfy.add.js` / `fog.ntfy.list.js`.

### New: auth for protected topics
The unused `credentials` column is now wired up — `user:pass` → HTTP basic auth, anything else → `Authorization: Bearer <token>`. Empty = public topic.

### Verified so far
`php -l` clean across all files; full app bootstraps with the plugin (all 14 classes autoload); `schema()` emits valid non-destructive SQL; ntfy protocol confirmed against `ntfy.sh`. Still to do (manual, in-browser): install → table create, menu item, add-topic round-trip, and a live event firing.

Credit stays with you in the file headers. Once it's verified in the UI I'll open a PR from `feature-760-ntfy` into `working-1.6`.

mastacontrola:
### Landed on `working-1.6` 🎉

This is now merged into `working-1.6` as **#842** (merge `9ac9b9a9`). Your authorship is preserved in the new file headers — thanks again, @TonyKaito.

Rather than merge here against `dev-branch`, we moved it to `working-1.6`, which is where notification plugins are heading and which recently gained a **per-plugin schema-migration system** any new table-owning plugin must use. How we approached the port:

**Brought to `working-1.6` conventions**
- Schema via `createSql()` + append-only `schema()` + non-destructive `install()` (`Schema::applyUpdates()`), so re-install no longer drops data and upgrades flow through the dashboard "update available" banner. See `docs/PLUGIN_SCHEMA_MIGRATIONS.md`.
- Added the missing **API** + **JS** hooks (the JS hook is why `fog.ntfy.*.js` loads); menu hook moved to the `hook_main` pattern; page rebuilt on the `makeInput/formFields/makeFormTag` helpers with `checkAuthAndCSRF()` + `HTTPResponseCodes`; JS split into `add`/`list`.
- Page renamed to `ntfymanagement.page.php` / class `NtfyManagement` so the page loader (which only registers `*.page.php` whose class matches the filename) picks it up.

**Correctness fixes**
- The handler now speaks **ntfy's** protocol (message = POST body, title = `Title` header) instead of pushbullet's JSON, so notifications actually render. Verified against ntfy.sh.
- Wired up the previously-unused `credentials` column for protected topics: `user:pass` → basic auth, else `Authorization: Bearer <token>`.
- Fixed `NtfyException` typo, bareword array keys (PHP 8 fatal), the undefined `_apiKey` path, an unescaped form value, and a dedup that hit a core `distinct()` multi-placeholder bug (now uses `exists()`).

**Re: the event closure / `dev-branch`**
While testing, the fired-event path 500'd on `working-1.6` because `self::$eventloop($x)` is parsed as a static method named by a local variable, not as invoking the closure. Worth noting for context here: **`dev-branch` is _not_ affected** — its pushbullet (and your original PR) invoke via `array_map(self::$eventloop, ...->find())`, which passes the closure as a value. The bug is specific to the `working-1.6` `foreach { self::$eventloop($x) }` refactor (which also swapped `find()` for `Route::listem`, since `working-1.6` dropped `find()`). #842 fixes it for both ntfy and the existing pushbullet plugin on `working-1.6`; nothing to change on `dev-branch`.

Since the feature now lives on `working-1.6`, this PR can be closed unless you'd also like a `dev-branch` build of it.

mastacontrola:
The ntfy notification plugin has been ported to and merged into working-1.6 via #842, so this functionality ships with FOG 1.6. Closing this PR against dev-branch — anything still wanted on the 1.5.x line can be back-ported from working-1.6, but the recommendation is to pick it up in 1.6. Thanks again for the contribution!

PR #767, opened 2025-12-30
Update functions.sh for Debian Trixie installation - PHP 8.4

pemaier:
php-mysqlnd removed because it is starting with PHP 8.1 only a virtual package. It is included in php-mysql.


pemaier:
Sorry wrong branch..

PR #768, opened 2025-12-30
Update functions.sh for Debian Trixie installation - PHP 8.4

pemaier:
php-mysqlnd removed because it is starting with PHP 8.1 only a virtual package. It is included in php-mysql.

### DIFF
--- a/lib/common/functions.sh
+++ b/lib/common/functions.sh
@@ -845,7 +845,7 @@ installPackages() {
                 done
                 ;;
             php-mysql*)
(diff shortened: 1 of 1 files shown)


mastacontrola:
Is this fixing something or just something you were cleaning up?

The code is built such that if php-mysqlnd isn't found, it will simply fall back to php-mysql (so in either case would perform the same thing?)

And while what you're looking at is true for Debian Trixy and php > 8.1, what about those users who aren't that updated or not using php 8 at all yet?

Neustradamus:
@pemaier: Any progress on it?

PR #780, opened 2026-02-06
Fix Power Management "Perform Immediately" bug.

jpugliese:
When pmOndemand=1, the backend still requires pmMin/pmHour/etc. This patch populates time fields before save(), preventing "Required database field is empty: min".

Tested on FOG 1.5.x with shutdown and reboot tasks.

### DIFF
--- a/packages/web/lib/fog/powermanagement.class.php
+++ b/packages/web/lib/fog/powermanagement.class.php
@@ -110,6 +110,13 @@ class PowerManagement extends FOGController
      */
     public function save()
     {
+        if ($this->get('pmOndemand') == '1') {
+			$this->set('pmMin', date('i'));
+			$this->set('pmHour', date('H'));
+			$this->set('pmDom', date('d'));
+			$this->set('pmMonth', date('m'));
+			$this->set('pmDow', date('w'));
+		}
         parent::save();
         return $this
             ->assocSetter('PowerManagement', 'host', true)


darksidemilk:
Thank you for your contribution!
As soon as we have a chance to test this internally we'll get it added!

mastacontrola:
So I have added code that should prevent the need to do this. Your method, while a work around doesn't actually fix the real problem. Your method effectively makes the Power On task a yearly task (whenever the dow matches the dom). (So a ~ 4 year automated task.)

Basically the problem is we added coded to prevent updating things if a "required" database field isn't defined. This has some unexpected side effects while we fix the underlying code appropriately.

Please review with the latest code in dev-branch (not your patch please) and validate that Powermanagement Ondemand items works appropriately now?

Thank you and apologies on the delay.

Neustradamus:
@jpugliese: Any progress on it?

mastacontrola:
Thanks for the report and the patch — the underlying problem was real, but this workaround sets the raw DB column names (`pmMin`, `pmHour`, …) through the ORM, which only accepts the friendly field names, so the values never actually reach the database.

The proper fix landed on dev-branch in 229a047bc: `PowerManagement` no longer requires the cron fields (min/hour/dom/month/dow), since on-demand tasks legitimately have none. The same fix (plus additional on-demand handling repairs) has now also been applied to working-1.6.

Since the fix has been available for a while with no further response, closing this PR. If you still see on-demand power management failures on a current dev-branch build, please open an issue.

PR #782, opened 2026-02-11
Update Portuguese translation

eduardomozart:
Removed duplicate strings from source code and updated/fixed Portuguese translation.


darksidemilk:
This looks pretty straight forward and only effecting ui.

darksidemilk:
We do need to port it all into 1.6 too

PR #784, opened 2026-02-12
Change input sanitization to use FILTER_SANITIZE_SPECIAL_CHARS

eduardomozart:
Fix error in PHP error log on PHP 8.3:

```
PHP Deprecated: Constant FILTER_SANITIZE_STRING is deprecated
```

### DIFF
--- a/packages/web/commons/init.php
+++ b/packages/web/commons/init.php
@@ -22,7 +22,7 @@ class Initiator
         if (!self::$sanitizeItems) {
             self::$sanitizeItems = function (&$val, $key) {
                 if (is_string($val)) {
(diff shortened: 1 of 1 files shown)


darksidemilk:
Thank you for your contribution! We probably need to do this in some sort of try catch with all the php versions that could be out in the wild.

eduardomozart:
This is compatible with PHP 5 onwards, which is already EOL at this stage.

darksidemilk:
Well then....

osiktech:
May I suggest, to also merge this commit into working-1.6 branch? There are people in the wild using the working-1.6 branch 😉

PR #788, opened 2026-02-16
Update snapinmanagementpage.class.php

Gauthier-L:
Add Fix for "Snapin Run With Args", because Fog Client won't run without this fix..


mastacontrola:
Okay so this is built against the stable branch when it should be the dev-branch. I have addresses this as well as decoding the entered rw code within the database as required. Closing. Thank you for the assist though as I did forget to account for the runWith side of things.

PR #802, opened 2026-03-24
Persistent Groups Manager does not auto create snapin tasks, only assigns the snapins.

PorygonPhD:
In a test deployment, I noticed that we had to deploy all the snapins through the GUI even though the snapins were assigned to each new host. 

I dug further and found a forum post regarding that same issue that had some sample code written for it.

https://forums.fogproject.org/topic/15084/persistent-groups-snapins-added-to-host-but-not-deployed/33?lang=en-US&page=2

I refined it and implemented it into my current FOG installation that I've been testing out and it has worked on ~30 workstations so far.

I'm new to Git but wanted to hopefully help others that had the same issue. If I need to submit an issue with this, please let me know and I can give more details if needed.

### DIFF
--- a/packages/web/commons/init.php
+++ b/packages/web/commons/init.php
@@ -20,15 +20,22 @@ class Initiator
     private static function setSanitize(): void
     {
         if (!self::$sanitizeItems) {
-            self::$sanitizeItems = function (&$val, $key) {
-                if (is_string($val)) {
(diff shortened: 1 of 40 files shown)


PorygonPhD (commented):
Updated `persistentgroupsmanager.class.php`.

PR #809, opened 2026-03-27
Update translation for zh_CN.UTF-8

BA4TMR:
Here is the updated messages.po file as requested.


mastacontrola:
Alright, while i have merged this, there appears to be texts that you had that no longer exist, and there appears you are missing texts that were added.

I've rerun the to try to capture those things, if you don't mind reviewing?

I appreciate it and I'm sure the community will as well. Thank you!

BA4TMR:
Okay, which files are you missing?

PR #834, opened 2026-06-16
Update functions.sh

Nono-m0le:
Check httpbin.org before neverssl.com, as this 2nd one doesn't seems to be reachable


mastacontrola (approved):
Sound change: `checkInternetConnection` breaks on the first reachable host, so ordering the more reliable `httpbin.org` ahead of the flaky `neverssl.com` makes the connectivity check succeed faster and more consistently. Retargeted from stable to dev-branch. Thanks @Nono-m0le!

PR #836, opened 2026-06-16
Fix for Full Manual Registration - fails unless a product key is entered even though it says optional. This allows user to hit enter and submit with blank entry.

wwtechnologies:


mastacontrola:
Thanks very much for catching this and submitting the fix, @wwtechnologies! 🙏

You're right — the `{1,29}` regex required at least one character, so a blank product key (which is supposed to be optional) failed validation and threw "Invalid product key supplied". Good find.

A couple of notes on what I did:
- I retargeted this from `stable` to `dev-branch` — `stable` is auto-updated downstream, so fixes land on `dev-branch` first.
- Merged into `dev-branch` (squash) crediting you.
- I also ported the same fix to `working-1.6`, since that branch has the equivalent check and the same bug.

When you have a chance, could you pull `dev-branch` (and `working-1.6` if you're testing there) and confirm Full Manual Registration now submits cleanly with a blank product key, and still rejects a malformed one? Appreciate the help!

PR #837, opened 2026-06-18
Fix rfc2307 LDAP group login

fussinator:
Admin/Mobile LDAP group access is broken since commit 1c5e04ddb891b25716e59ea5d65abde39704d562.

RFC2037's posixGroup uses simple "memberUid: johndoe" attribute/value pairs, but the code is trying to filter groups using "cn: johndoe".  In otherwords, $usrNamAttr has been used instead of $grpMemAttr.

The resulting filter string is something like:
(&(|(name=fogadmins)(memberuid=fogadmins))(|(memberuid=cn=johndoe,ou=accounts,dc=example,dc=org)(memberuid=cn=johndoe)(cn=johndoe)))

Instead of the intended:
(&(|(name=fogadmins)(memberuid=fogadmins))(|(memberuid=cn=johndoe,ou=accounts,dc=example,dc=org)(memberuid=cn=johndoe)(memberuid=johndoe)))

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,203 @@
+# Interaction rules (apply in every conversation)
+
+1. Don't assume. Don't hide confusion. Surface tradeoffs.
+2. Minimum code that solves the problem. Nothing speculative.
+3. Touch only what you must. Clean up only your own mess.
+4. Define success criteria. Loop until verified.
+
+---
+
+# FOG Project — Claude Context (dev-branch / 1.5.x)
+
+> **Branch note:** This is `dev-branch`, the **1.5.x stable/patches line** (currently `1.5.10`).
+> It is a *different codebase* from `working-1.6` — file naming, layout, and some conventions
+> differ (see below). Do **not** assume 1.6 patterns or file paths apply here. Fixes often need
+> to be ported between the two branches by hand, not merged.
+
+## What This Is
+
+FOG Project is an open-source network imaging and endpoint management system. It allows IT administrators to deploy OS images to computers over PXE boot, manage client hosts, schedule tasks, push snapins (software packages), manage printers, track users, and run background replication/multicast services.
+
+- Primary language: **PHP** (the codebase runs on PHP 8; `commons/init.php` uses typed signatures like `Initiator::e(mixed $value): string`)
+- Frontend: **jQuery + Bootstrap + AdminLTE** (no build step — plain JS served directly)
+- Background services: **PHP CLI daemons**
+- ~128 core `*.class.php` files under `lib/fog/`
+- Version line: **1.5.10** (the pre-commit hook stamps `dev`/`stable` branches with the `Patches` channel)
+
+---
+
+## How this differs from `working-1.6` (read this first)
+
+| Thing | dev-branch (1.5.x) | working-1.6 |
+|------|--------------------|-------------|
+| Page file names | `imagemanagementpage.class.php` (one word, `.class.php`) | `ImageManagement.page.php` |
+| Page class names | `ImageManagementPage` (with `Page` suffix) | `ImageManagement` (no suffix) |
+| ORM array syntax | older `array(...)` in many files | `[...]` short syntax |
+| `declare(strict_types=1)` | **none** in `lib/fog/` | present in newer files |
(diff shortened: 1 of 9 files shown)


mastacontrola (commented):
Thanks for tracking this down, @fussinator — the diagnosis is exactly right. `$usrNamAttr` was being used where `$grpMemAttr` belongs, so RFC2307 `posixGroup` membership (`memberUid: johndoe`) never matched and the third filter alternative resolved to the nonsensical `(cn=johndoe)`. The fix itself is correct.

The one thing I need to ask before merging: **could you retarget this PR to `dev-branch` rather than `stable`?**

A couple of reasons:

- `stable` isn't the branch we take PRs against — it's downstream and gets updated from our development branches, so changes landed there directly get clobbered.
- Because the branch is based on `stable`, the diff has picked up a lot of changes that **already exist on `dev-branch`** and aren't really yours — `CLAUDE.md`, `lib/common/functions.sh`, the three `lib/*/config.sh`, `packages/web/service/progress.php`, the `registration.class.php` product-key tweak, and the `system.class.php` version bump. Once you retarget to `dev-branch`, all of that collapses away and the PR cleanly reduces to just the LDAP fix, which is what we want to review and merge.

I've confirmed `ldap.class.php` is identical between `stable` and `dev-branch`, so your fix applies as-is there with no conflicts.

Heads up for visibility: the 1.6 line (`working-1.6`) carries a rewritten version of this plugin where the same bug exists in two spots, so I'll port an equivalent fix there separately — no action needed from you on that.

One related thing I found while reviewing this code path, in case you want to fold it into this PR (otherwise I'll handle it separately): a few lines down, the fallback membership check builds a regex straight from the user DN —

```php
$pat = sprintf('#%s#i', $userDN);
...
$adm = preg_grep($pat, $users);
```

Because `$userDN` isn't run through `preg_quote()`, a DN containing regex metacharacters (`.`, `(`, `+`, `*`, `#`, etc.) either matches too loosely or produces an invalid pattern, in which case `preg_grep()` returns `false` and that fallback path silently denies access. Wrapping it as `preg_quote($userDN, '#')` fixes it. Not required for this PR — just flagging since it's in the same function.

Thanks again!

mastacontrola:
Thanks again for this, @fussinator — the diagnosis and fix were spot on.

Rather than have you rebase/retarget by hand, I've carried your fix over to **#839** against `dev-branch` with your original commit and authorship preserved. While I was in that same code path I also folded in a couple of related fixes I found:

- escaping the user DN with `preg_quote()` before it's used as a regex in the fallback membership loop (an unescaped DN with regex metacharacters could silently deny access), and
- reworking `checkAddUser()` so that with multiple LDAP servers configured, a server where the user is absent can no longer clobber a valid match from another server (it now keeps the most privileged result instead of acting per-server).

So I'm going to close this one in favour of #839 — but the rfc2307 commit there is yours. Equivalent fixes are also going into the 1.6 line in #838. Really appreciate you tracking this down and writing it up so clearly.

fussinator:
On 6/18/26 4:45 AM, Tom Elliott wrote:
> *mastacontrola* left a comment (FOGProject/fogproject#837) 
> <https://github.com/FOGProject/fogproject/pull/837#issuecomment-4741160079>
>
> Thanks again for this, @fussinator <https://github.com/fussinator> — 
> the diagnosis and fix were spot on.
>
> Rather than have you rebase/retarget by hand, I've carried your fix 
> over to *#839 <https://github.com/FOGProject/fogproject/pull/839>* 
> against |dev-branch| with your original commit and authorship preserved.
>

Doh!  Thank you for cherry picking that, I'm sorry I missed that I based 
off the wrong branch.

There's something else bugging me in that filter string that I haven't 
had time to verify yet.  It seems the first sub-expression filters 
objects on a name that matches the admin/mobile group (e.g. 
name=fogadmin) or a ${grpMemAttr}=<admin/mobile group> (e.g. 
memberUid=fogadmin in my case).

It is highly unlikely that the name of the admin/mobile group will ever 
be a member name of another group, unless perhaps the admin group 
happens to be a single-user user-group pairing group, or it's somehow 
needed for nested groups.

Is there another reason I don't understand for that first sub-expression?

I'm out of town for the next few days, so I probably won't get back to 
this until next week.

PR #870, opened 2026-07-15
Add Japanese translation

mashiro-mikage:
## Summary

This adds Japanese translation support for the FOG web UI.

Changes include:

- Add `ja_JP.UTF-8` translation files
- Add Japanese to the web UI language selector
- Add `ja` / `JP` locale handling
- Install Japanese language packs when internationalization support is enabled

## Testing

- Installed FOG on Ubuntu with internationalization enabled
- Confirmed `ja_JP.UTF-8` locale support is installed
- Confirmed Japanese is available in the language selector
- Confirmed the web UI renders in Japanese for the main pages and common workflows

## Notes

Some dynamically generated labels may still appear in English if they are not passed through gettext yet. Those can be handled separately as follow-up i18n fixes.


mastacontrola:
Thank you so much for this, @mashiro-mikage! 🎉 This is a genuinely high-quality contribution — the catalog is complete (2284 translated messages), compiles cleanly under `msgfmt --check` with zero errors or fuzzy entries, and everything follows FOG's existing i18n conventions exactly (locale registration in `Initiator::language()`, the selector label in `text.php`, and both the `glibc-langpack-ja` / `language-pack-ja` installer hooks). Merged into `dev-branch`. 🙏

For your reference, here's what happened to it downstream:

**Ported to `working-1.6`** (commit for the 1.6 line): The 1.6 branch is a partially divergent codebase, so this wasn't a straight cherry-pick — a few notes in case you want to contribute there directly in the future:
- `text.php` uses `$foglang['Language']` on 1.6 (vs `$foglangt['Language']` on dev-branch).
- The installer (`lib/common/functions.sh`) has **two** package functions on 1.6 (`listPackages()` and `installPackages()`), so the language list needed the `ja` entry added in four places rather than two.
- Your `ja_JP.UTF-8/messages.po` was translated against the dev-branch string set. When ported to 1.6, our pre-commit hook runs `msgmerge` against 1.6's `messages.pot`, which reconciled it to **1262 translated / 809 fuzzy / 232 untranslated**. The drop is purely because 1.6's source strings have drifted from 1.5.x — the fuzzy/untranslated ones fall back to English at runtime, exactly like the other languages. If you're ever up for a follow-up pass to un-fuzzy the near-matches on the 1.6 catalog, it'd be very welcome, but there's no obligation — this is already a great addition as-is.

As you noted, some dynamically generated labels that aren't yet passed through gettext will still show in English; those are a separate i18n cleanup and not a gap in your work.

Thank you again for taking the time to do this properly — 日本語 support is a real win for the project. 🙌

PR #906, opened 2026-07-27
Add isPopulated() so association guards stop trusting isLoaded()

darksidemilk:
## Summary
- `FOGBase::isLoaded($key)` is a test-and-set recursion guard for the internal `get()`/`set()`/`loadItem()` lazy-load chain — every call marks the key loaded, even one that answers `false`. That's required internally (a `loadX()` method's own `set()` call must not see "not loaded" again and recurse forever), but it means a bare `if ($this->isLoaded($key))` used as a standalone precondition elsewhere silently poisons the flag for the rest of the request.
- That's exactly how this went wrong: `Host::save()`'s snapins guard (added in 8e31b5cf0 to strip stale/blank snapin ids) checked `isLoaded('snapins')` on a plain FOG client check-in where nothing else touches `'snapins'` first. The check answered `false` (correctly — guard skipped) but poisoned the flag anyway. `assocSetter('Snapin', 'snapin')`, called right after, then trusted the poisoned flag and proceeded on `get('snapins')`'s empty-string fallback.
- **Severity correction (h/t review comment from a parallel investigation):** this isn't just a phantom-row-insertion bug. `assocSetter()`'s removal half runs *before* its insert half and had no guard of its own: `(array)''` casts to `['']`, and `array_diff($cur, [''])` subtracts nothing, so `$rem` came out as the host's **entire real snapin list**. On the first check-in a host received after picking up a build containing 8e31b5cf0 (2026-06-24 onward), every one of its real `snapinAssoc` rows was deleted, and only then did a single phantom `saSnapinID = 0` row get inserted. Subsequent check-ins just churned that 0 row (delete + reinsert), which is what steady-state production logging showed. **A `DELETE FROM snapinAssoc WHERE saSnapinID = 0` cleanup does not restore the lost real assignments** — any host that checked in during that window needs its snapins manually reassigned. Worth flagging in release notes since the failure was silent (no error, no log line, Snapins tab just came back empty).
- Rather than patch each guard individually (an earlier version of this PR did exactly that), this fixes the general mechanism: added `FOGBase::isPopulated($key)` — `isset($this->data[$key])`, matching `get()`'s own fallback condition exactly, so it only reports true when `get()` is actually about to return real data — and switched every standalone gate to it:
  - `assocSetter()`'s own "don't work on an unloaded item" check — the single shared implementation behind **all 24 call sites across every association type** (groups, modules, printers, snapins, role users/usergroups, storage groups, powermanagement, and the OU/WindowsKey/Location/Site/LDAP plugins). This is the high-leverage fix: no bare `isLoaded()` check anywhere else, present or future, can fool `assocSetter()` again, and the delete-everything failure mode above is now impossible for any of the 24 relations, not just snapins.
  - `Host::save()`'s snapins guard.
  - `Role::_syncPermissions()` — identical shape: a role save that never touches permissions (e.g. a plain rename) would poison the flag, so a later `get('permissions')` on that object (e.g. building the API response after `save()`) would see an empty list instead of the role's real permissions.
  - `Image::save()`'s hosts-reassignment guard — also fixes a latent second-save `TypeError` (`count()` on a non-countable `''`) that a parallel review flagged for the pre-isPopulated() version of this guard.
- `isLoaded()` itself, and the internal `get()`/`set()`/`add()`/`remove()` call sites that pair it with an immediate load, are unchanged — recursion-safety isn't touched.
- Also added a defensive fallback in `assocSetter()`: guard `$items` against a falsy (e.g. empty-string) `get()` result before the `(array)` cast, so a bad value can't turn into `['']` and get diffed in as a phantom association or a mass-delete — belt and suspenders alongside the isPopulated() fix, and now provably unreachable via the poisoning path since the isPopulated() gate returns early first.

## Test plan
- [ ] Deploy to a test FOG server and let a host check in repeatedly; confirm no new `snapinAssoc` rows with `saSnapinID = 0` appear, and confirm existing real snapin assignments survive repeated check-ins.
- [ ] Edit a role's name without touching permissions, then re-fetch it via the API; confirm `permissions` still reflects the real list.
- [ ] Edit a role's permissions and confirm they still save correctly (isPopulated() gate doesn't regress the working case).
- [ ] **Remediation on affected installs is two steps, not one:** (1) `DELETE FROM snapinAssoc WHERE saSnapinID = 0` to clear the phantom rows; (2) any host that checked in between 2026-06-24 and this fix landing needs its real snapin assignments manually recreated — they were deleted, not just polluted.

A companion PR applies the equivalent fix to `dev-branch` (which has its own independent set of bare `isLoaded()` gates — different call sites, same underlying bug shape — but never received 8e31b5cf0 and was never vulnerable to the specific snapin data-loss bug this PR was originally opened for).

### DIFF
--- a/packages/web/lib/fog/fogbase.class.php
+++ b/packages/web/lib/fog/fogbase.class.php
@@ -1084,7 +1084,25 @@ abstract class FOGBase
         return -1;
     }
     /**
-     * Check if isLoaded.
+     * Internal recursion guard for the get()/set()/loadItem() lazy-load
+     * chain -- NOT a "does this key hold data" predicate.
+     *
+     * This is a test-and-set: every call marks $key loaded, even one that
+     * returns false. That side effect is required so a loadX() method's
+     * own set() call doesn't see "not loaded" again, re-trigger loadItem(),
+     * and recurse forever. It means a bare `if ($this->isLoaded($key))`
+     * used as a standalone precondition -- anywhere the false branch isn't
+     * immediately followed, in the same call, by an actual load -- will
+     * silently poison the flag for the rest of the request: later code
+     * that calls get($key) expecting a real lazy-loaded value instead
+     * sees the (now-true) flag, skips the real load, and falls back to
+     * get()'s "never set" default.
+     *
+     * Use isPopulated($key) instead for "do I actually have data for this
+     * key" checks -- it has no such side effect. This exact confusion
+     * (Host::save()'s snapins guard + assocSetter()'s own isLoaded() gate)
+     * caused phantom saSnapinID=0 rows to be inserted on every FOG client
+     * check-in.
      *
      * @param string|int $key the key to see if loaded
      *
@@ -1097,6 +1115,23 @@ abstract class FOGBase
         $this->isLoaded[$key] = true;
 
         return $result ? $result : false;
+    }
+    /**
+     * Whether $key currently holds resolved data -- a pure,
+     * side-effect-free predicate. Unlike isLoaded() (a recursion guard for
+     * the internal lazy-load chain; see its docblock), this is safe to use
+     * as a standalone precondition anywhere the caller means "do I have
+     * real data for this key," e.g. "only sync this association if the
+     * caller actually touched it."
+     *
+     * @param string|int $key the key to check
+     *
+     * @return bool
+     */
+    protected function isPopulated($key)
+    {
+        $key = $this->key($key);
+        return array_key_exists($key, $this->data);
     }
     /**
      * Reset request variables.
--- a/packages/web/lib/fog/fogbase.class.php
+++ b/packages/web/lib/fog/fogbase.class.php
@@ -1124,6 +1124,13 @@ abstract class FOGBase
      * real data for this key," e.g. "only sync this association if the
      * caller actually touched it."
      *
+     * Deliberately isset(), not array_key_exists(): get()'s own fallback
+     * to '' is gated the same way (`isset($this->data[$key]) ? ... : ''`),
+     * so this only reports true when get() is actually about to return
+     * real data instead of that fallback -- the two must agree, or a key
+     * explicitly set to null would read as "populated" here while get()
+     * still silently handed back ''.
+     *
      * @param string|int $key the key to check
      *
      * @return bool
@@ -1131,7 +1138,7 @@ abstract class FOGBase
     protected function isPopulated($key)
     {
         $key = $this->key($key);
-        return array_key_exists($key, $this->data);
+        return isset($this->data[$key]);
     }
     /**
      * Reset request variables.
--- a/packages/web/lib/fog/fogcontroller.class.php
+++ b/packages/web/lib/fog/fogcontroller.class.php
@@ -1263,13 +1263,23 @@ abstract class FOGController extends FOGBase
         $objstr = "{$obj}ID";
         $assocstr = "{$alterItem}ID";
 
-        // Don't work on item that isn't loaded yet.
-        if (!$this->isLoaded($plural)) {
+        // Don't work on item that isn't populated yet. isPopulated(), not
+        // isLoaded(): isLoaded() marks a key loaded on every call, even
+        // one that answers false, so a bare isLoaded() gate anywhere else
+        // in the request (present or future) could poison this check and
+        // make it proceed on stale/empty data. isPopulated() only reports
+        // true once real data is actually here.
+        if (!$this->isPopulated($plural)) {
             return $this;
         }
 
-        // Get the current items.
+        // Get the current items. Guard against a non-array fallback (e.g.
+        // an empty string) so a bad value can't cast to [''] below and be
+        // diffed in as a phantom "new" association.
         $items = $this->get($plural);
+        if (!$items) {
+            $items = [];
+        }
 
         // Fetch current associations from the database.
         $cur = Route::getIds(
--- a/packages/web/lib/fog/host.class.php
+++ b/packages/web/lib/fog/host.class.php
@@ -282,19 +282,20 @@ class Host extends FOGController
         // the "< 1" guard on setSnapinOrder()'s save path; without this a 0
         // would be inserted and later surface as a phantom run-order entry.
         //
-        // Deliberately unconditional: FOGBase::isLoaded() marks a key loaded
-        // on every call, even when it answers false and nothing follows up
-        // by loading it. Gating this on isLoaded('snapins') let a plain
-        // check-in (nothing else touches 'snapins' first) poison that flag,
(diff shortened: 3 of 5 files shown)


mastacontrola:
Confirmed independently from a static read of the same code paths — the diagnosis here is correct, and the production instrumentation is better evidence than anything I derived. This should land.

Two additions, one of which materially changes the remediation advice for affected installs.

## The delete half — affected installs have lost real snapin assignments, not just gained junk rows

The PR characterizes the symptom as a phantom `saSnapinID=0` row being inserted. That's real, but it's the *second* thing that happens. `assocSetter`'s removal half runs first and has no guard on it (`fogcontroller.class.php:1282-1291`):

```php
$items = $this->get($plural);              // '' — poisoned flag, load skipped
$cur   = Route::getIds($classCall, ...);   // the host's real snapin ids
$rem   = array_diff((array)$cur, (array)$items);
if (count($rem ?: [])) {
    Route::deletemass($classCall, [$assocstr => $rem, $objstr => $this->get('id')]);
}
```

`(array)''` is `['']`, which subtracts nothing, so `$rem` is the host's **entire** current association list:

```php
php > var_dump(array_diff(["3","7"], (array)""));
array(2) { [0]=> string(1) "3" [1]=> string(1) "7" }
```

So on the first check-in after a host picks up a build containing 8e31b5cf0, every one of its `snapinAssoc` rows is deleted, and then a single `saSnapinID=0` row is inserted. Subsequent check-ins just churn that 0 row (delete it, re-insert it), which is consistent with the steady-state you observed.

This means the cleanup in the test plan is correct but incomplete:

> Existing phantom `saSnapinID = 0` rows are pre-existing data and are not touched by this change — needs a one-time `DELETE FROM snapinAssoc WHERE saSnapinID = 0` cleanup on affected installs.

That's right, but no cleanup can restore what was deleted. Anyone who ran a working-1.6 build from 2026-06-24 onward and had hosts check in has lost their host→snapin assignments and needs to reassign them. Worth calling out explicitly in the release notes, since the failure is silent — no error, no log line, and the Snapins tab just comes back empty.

## Scope — the gate is fixed at 2 call sites, the mechanism is intact at the other 22

Fully agree with not speculatively rewriting working code, and I don't think any of this should block merging. Recording it so it isn't lost:

`assocSetter()` is called 24 times across 15 core classes and 4 plugins (hosts, groups, modules, printers, users, usergroups, roles, snapins, storagegroups, powermanagement, plus ldap/location/ou/site/windowskey). Its own guard at `fogcontroller.class.php:1267` is still `if (!$this->isLoaded($plural))` — still test-and-set — so `Host::save()` after this PR still does:

- `assocSetter('Group','group')` → first `isLoaded('groups')` answers `false`, returns early, **marks `groups`**
- from then on, on that instance, `$host->get('groups')` returns `''` rather than the group list
- a second `save()` on that instance → the guard now passes → `$items === ''` → every group association deleted, `groupID=0` inserted

Same for `modules` and `printers`. That's the identical failure this PR fixes for `permissions`, and the justification given for the `role.class.php` change ("any later `get('permissions')` on this same object ... would see an empty list instead of the role's real permissions") applies verbatim to `get('groups')`.

In fairness to the current scope, that variant is latent rather than live: the REST update path re-instantiates before serializing (`route.class.php:2436`, `$class = new $class($id)`), so the API response isn't affected. It needs a same-instance re-read or a double save.

Two smaller ones:

- **`image.class.php:132`** — agreed it isn't exploitable on a single save. On a second save of the same instance it's `count($this->get('hosts'))` where `get()` returns `''`, which under PHP 8 is an uncaught `TypeError`, so the failure mode there is a fatal rather than a wrong result.
- **The 0-id insert is still general.** `assocSetter`'s insert loop (`:1311-1322`) writes `$diff` verbatim with no validation, for every relation. This PR sanitizes host↔snapin only, and only as a side effect of 8e31b5cf0's filter now being unconditional. A blank id reaching any of the other 23 relations still inserts a 0 row.

I'm putting up a follow-up PR that addresses the mechanism rather than the call sites — an explicit dirty-key set on `FOGController` populated only by `set()`/`add()`/`remove()`, so `assocSetter` asks "did the caller give me a list?" instead of "has anyone touched this key?", plus moving the positive-id filter into `assocSetter` so it covers all 24 relations at the right layer. It'll build on this, not replace it. dev-branch carries the same `assocSetter` guard (`fogcontroller.class.php:1150`) plus a private copy in the site plugin, so it needs the port too.

darksidemilk:
Thanks for the deep-dive — the delete-before-insert finding is a real severity correction and I've folded it into the PR description: `assocSetter()`'s removal half ran unguarded, so `(array)'' === ['']` meant `$rem` came out as the host's *entire* real snapin list, not just an empty diff. Updated the summary and test plan to make clear the fix on affected installs is two steps (clear the phantom 0-rows *and* manually reassign snapins to any host that checked in from 2026-06-24 onward), not one.

Since your comment, I'd already pushed a redesign that changes the scope claim somewhat — worth reconciling before you finalize the follow-up:

- `assocSetter()`'s own gate (`fogcontroller.class.php:1267`) is no longer `isLoaded()` — it's switched to a new `FOGBase::isPopulated($key)` (`isset($this->data[$key])`, matching `get()`'s own fallback condition exactly). That's the shared implementation behind all 24 call sites, so the `groups`/`modules`/`printers` latent case you traced (poison on save #1, mass-delete-and-zero on save #2 of the same instance) should already be closed, not just the `permissions`/`snapins`/`hosts` ones. Same for the `Image::save()` second-save `TypeError` you flagged — `isPopulated()` reporting true now means `get()` is guaranteed to return the real value, not `''`, so that path shouldn't fire either. Would appreciate you double-checking that trace against the current PR head (`1b1d6ad4b`) since I'd rather have a second reader confirm than assume.
- That said, your dirty-key-set idea is a meaningfully different (and in one respect better) design than `isPopulated()`: a set populated only by `set()`/`add()`/`remove()` would correctly skip `assocSetter()`'s work entirely when a key was only *read* (lazy-loaded for display) and never actually changed, where `isPopulated()` will harmlessly re-diff against identical data. Not a correctness gap for us — the diff comes out empty either way — but it's a real precision difference worth keeping in your follow-up if the extra query volume matters at scale.
- Fully agree the "0-id insert is still general" point stands and isn't addressed here — `assocSetter()`'s insert loop still writes `$diff` verbatim with no per-element validation, for every relation. This PR closes the *flag-poisoning* path that let bad values reach that loop; it doesn't add input sanitization at the assocSetter layer itself. That's real additive hardening and I'd rather it land as your planned follow-up than have me duplicate scope here.

Porting the equivalent fix to `dev-branch` now (separate worktree/branch, per its own independent set of bare `isLoaded()` gates) — will flag here once that PR is up so the two can be reviewed together.

PR #911, opened 2026-07-27
Add isPopulated() so association guards stop trusting isLoaded()

darksidemilk:
## Summary
Companion to #906 (working-1.6). `FOGBase::isLoaded($key)` is a test-and-set recursion guard for the internal `get()`/`set()`/`loadItem()` lazy-load chain — every call marks the key loaded, even one that answers `false`. That's required internally, but it means a bare `if ($this->isLoaded($key))` used as a standalone precondition elsewhere silently poisons the flag for the rest of the request: later code calling `get($key)` expecting a real lazy-loaded value instead sees the (now-true) flag, skips the real load, and falls back to `get()`'s `''` default — which, cast to an array, becomes `['']` and gets diffed in as a phantom association, or worse, makes `assocSetter()`'s unguarded removal half treat the object's *entire real association list* as stale and delete it.

`dev-branch` never received the working-1.6 commit that introduced the specific snapins guard behind #906 (and persists host↔snapin associations through an entirely different, older mechanism, so it was never vulnerable to that specific bug). But it carries its own independent set of bare `isLoaded()` gates with the identical shape:
- `assocSetter()`'s own "don't work on an unloaded item" check (`fogcontroller.class.php:1150`) — the shared implementation behind every core association type on this branch.
- The Site plugin's own duplicated copy of `assocSetter()` (`site.class.php:258`) — this branch hasn't consolidated it into the shared method the way working-1.6 has.
- `Host::save()`'s `mac`, `additionalMACs`, `pendingMACs`, and `powermanagementtasks` guards.
- `Image::save()`'s hosts-reassignment guard.

None of these were confirmed live-exploited on this branch (unlike the working-1.6 case, nothing else in the request currently poisons any of these specific flags before their own guard checks them), but they're all the identical footgun shape, and the shared `assocSetter()` gate in particular is a single choke point worth closing defensively rather than waiting for some future change to trip it the way 8e31b5cf0 did on working-1.6.

Fix, ported directly from #906: added `FOGBase::isPopulated($key)` — `isset($this->data[$key])`, matching `get()`'s own fallback condition exactly — and switched every standalone gate listed above to it. `isLoaded()` itself is unchanged, as are the internal `get()`/`set()`/`add()`/`remove()` call sites that pair it with an immediate load, and the already-correct `MACAddressAssociation::getHost()`/`Inventory::getHost()` pattern (check paired with an immediate load in the same block — not a bug, left alone).

Also added the same defensive fallback to the Site plugin's `assocSetter` copy that the shared one already had on this branch: guard `$items` against a falsy `get()` result before the `(array)` cast.

## Test plan
- [ ] Save a host without touching MACs/power-management tasks; confirm no unintended MAC/power-management changes.
- [ ] Save an image without touching its host list; confirm no unintended host reassignment, and no `count()` `TypeError` on a repeated save of the same instance.
- [ ] Save a Site-plugin-managed host/user/group association without touching that specific relation; confirm no unintended deletion.

### DIFF
--- a/packages/web/lib/fog/fogbase.class.php
+++ b/packages/web/lib/fog/fogbase.class.php
@@ -960,7 +960,25 @@ abstract class FOGBase
         return -1;
     }
     /**
-     * Check if isLoaded.
+     * Internal recursion guard for the get()/set()/loadItem() lazy-load
+     * chain -- NOT a "does this key hold data" predicate.
+     *
+     * This is a test-and-set: every call marks $key loaded, even one that
+     * returns false. That side effect is required so a loadX() method's
+     * own set() call doesn't see "not loaded" again, re-trigger loadItem(),
+     * and recurse forever. It means a bare `if ($this->isLoaded($key))`
+     * used as a standalone precondition -- anywhere the false branch isn't
+     * immediately followed, in the same call, by an actual load -- will
+     * silently poison the flag for the rest of the request: later code
+     * that calls get($key) expecting a real lazy-loaded value instead
+     * sees the (now-true) flag, skips the real load, and falls back to
+     * get()'s "never set" default.
+     *
+     * Use isPopulated($key) instead for "do I actually have data for this
+     * key" checks -- it has no such side effect. This exact confusion, on
+     * the working-1.6 branch, caused phantom saSnapinID=0 rows (and, worse,
+     * mass-deletion of real associations) to be inserted on every FOG
+     * client check-in.
      *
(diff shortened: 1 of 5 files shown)


darksidemilk:
Superseded by #912, which covers this same isPopulated() fix plus the positiveIntIds() hardening from #910 and was empirically verified against real 1.5 lab servers on PHP 8.0/8.5. Closing to avoid a redundant/conflicting merge.

PR #921, opened 2026-07-27
ci: add PR-based backstop for stale FOG_VERSION/FOG_CHANNEL

darksidemilk:
## Summary

`packages/web/lib/fog/system.class.php`'s `FOG_VERSION`/`FOG_CHANNEL` are only ever bumped by the client-side `.githooks/pre-commit` hook. That hook never runs when a PR is merged via GitHub's web UI, so versions silently go stale on any PR merged that way — confirmed live before this change: `working-1.6` was 3 merges ahead of what its stored version reflected.

This adds `.github/workflows/check-fog-version.yml`, which on every push to `working-1.6`, `dev-branch`, `stable`, `rc-*`, or `feature-*`:
- Recomputes the expected version/channel using the **same formula** as the hook (duplicated independently in the workflow — `.githooks/pre-commit` itself is untouched).
- Compares it against what's actually committed.
- If they disagree, opens a **review-only PR** with the fix — it never pushes directly to a protected branch and never auto-merges.
- If they already match (hook or manual release process already did its job), it's a silent no-op.

This is a backstop, not a replacement — the existing local hook and the manual release/version-bump process for `stable` are unaffected, since every mechanism recomputes live from git state rather than relying on any cached value.

## Test plan
- [ ] Confirm the workflow runs on push to `working-1.6` and opens a PR proposing the correct caught-up version (currently `1.6.0-beta.2886` → should become `1.6.0-beta.2889`).
- [ ] Confirm a push to `dev-branch` produces its own separate PR.
- [ ] Confirm re-running after a fix PR is merged produces no further PR.
- [ ] Confirm no workflow run fires for branches outside the five watched patterns.

### DIFF
--- a/.github/workflows/check-fog-version.yml
+++ b/.github/workflows/check-fog-version.yml
@@ -1,0 +1,137 @@
+name: Check FOG version
+
+# Backstop for packages/web/lib/fog/system.class.php's FOG_VERSION/FOG_CHANNEL.
+#
+# The only thing that normally bumps those defines is .githooks/pre-commit,
+# which is a client-side git hook. GitHub's web-UI merge button never runs
+# repo-tracked hooks, so a PR merged that way leaves the version stale. This
+# workflow independently recomputes what the version should be (same formula
+# as the hook, duplicated here on purpose — see .githooks/pre-commit) and,
+# only if it disagrees with what's actually committed, opens a PR so a human
+# can review and merge it. It never pushes to a protected branch directly and
+# never auto-merges.
+
+on:
+  push:
+    branches:
+      - working-1.6
+      - dev-branch
+      - stable
+      - 'rc-*'
+      - 'feature-*'
+  workflow_dispatch: {}
+
+permissions:
+  contents: write
+  pull-requests: write
+
+concurrency:
+  group: fog-version-check-${{ github.ref_name }}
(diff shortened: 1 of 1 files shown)


darksidemilk:
Closing in favor of moving this logic out of `fogproject` and into `FOGProject/fog-workflows`, so it's owned in one place alongside `stable-releases.yml` (which already handles this repo's release automation from outside it) rather than duplicated per-repo.

The actual version-drift-detection logic (same formula as `.githooks/pre-commit`) now lives in fog-workflows as a reusable workflow: https://github.com/FOGProject/fog-workflows/blob/main/.github/workflows/check-fog-version.yml

`fogproject` only needs a thin push-triggered stub that delegates to it via `uses:` (GitHub Actions has no native cross-repo push trigger, so something has to live here to react to pushes/merges — this file should rarely need to change): #922 (`working-1.6`), #923 (`dev-branch`), #924/#925/#926 (existing `feature-*` branches).

Two differences from this PR's approach:
- `stable` is excluded — its version is owned entirely by `stable-releases.yml`'s release flow.
- Detected drift is committed directly to the branch instead of opening a review PR, since it's the same kind of mechanical fix the local hook already makes without review.

Thanks for flagging the underlying gap — the fix just moved to keep version-related automation in one place.


---

PR #925, opened 2026-07-28
ci: add push-triggered stub delegating to fog-workflows version check (feature-fog2-gui)

darksidemilk:
## Summary

Same stub as #922/#923, targeting this branch. Existing `feature-*`/`rc-*` branches don't automatically inherit the stub added to `dev-branch` (#923) since they predate it, so it needs to land here too for pushes on this branch to be watched — see #922 for the full rationale.

## Test plan
- [ ] Merge, then confirm a push to this branch triggers the fog-workflows reusable workflow and correctly computes/fixes drift if any exists.

---
_Generated by [Claude Code](https://claude.ai/code/session_013fKXqgACVAVHNpYm9ZXRwN)_

### DIFF
--- a/.github/workflows/check-fog-version.yml
+++ b/.github/workflows/check-fog-version.yml
@@ -1,0 +1,30 @@
+name: Check FOG version
+
+# Push-triggered trigger stub. All the actual version-drift-detection logic
+# lives in FOGProject/fog-workflows as a reusable workflow (workflow_call) -
+# this file exists only because GitHub Actions has no native cross-repo push
+# trigger, so something has to live here to react to pushes/merges. See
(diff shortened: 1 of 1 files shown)


darksidemilk:
Superseded — this stub was part of yesterday's first attempt at the FOG_VERSION check, which turned out to have a self-triggering commit loop bug (see the rewritten docs in FOGProject/fog-docs#72 for the full story). That design has been replaced with a single hourly-schedule workflow in fog-workflows that discovers watched branches itself via the GitHub API, so there's no stub file to propagate into fogproject at all anymore. Merging this would just add a dangling `uses:` reference to a workflow that no longer exists in fog-workflows.

Issue #960, opened 2026-08-01
Verify the Secure Boot chain on real hardware (shim to signed iPXE to signed FOS kernel)


mastacontrola:
**Is your feature request related to a problem? Please describe.**

#957 decided *what* FOG does about Secure Boot: nothing out of the box, opt-in and documented. The guide is merged at [fog-docs `secure-boot-signing`](https://github.com/FOGProject/fog-docs/blob/master/docs/kb/how-tos/secure-boot-signing.md).

What has **not** happened is anyone booting that chain on real hardware. Every individual link is upstream-supported and the artefacts have been inspected, but the end-to-end path is unproven and the guide says so in its Known limits. This issue tracks closing that gap.

**Describe the solution you'd like**

Boot a UEFI machine with Secure Boot **on**, through:

```
firmware  →  ipxe-shimx64.efi        (Microsoft UEFI CA 2011 + 2023, from ipxe/shim)
          →  ipxe.efi                (iPXE Secure Boot Intermediate G1A, extracted
                                      from ipxe-x86_64-sb.usb at EFI/BOOT/IPXE.EFI)
          →  autoexec.ipxe           (FOG's boot logic, served over TFTP)
          →  bzImage                 (signed locally, key enrolled via MOK)
```

and confirm it reaches the FOG menu and completes a deploy.

**Specific things to check, because they are the parts most likely to be wrong:**

1. **Where `autoexec.ipxe` has to live.** iPXE's `efi_autoexec_network()` calls `imgacquire()` with a bare filename, which resolves against **`cwuri`** — the directory the `.efi` was loaded from, not the TFTP root. So a stock `ipxe.efi` sitting at `/tftpboot/ipxe.efi` looks for `/tftpboot/autoexec.ipxe`, which FOG does **not** currently create; FOG only ships `/tftpboot/autoexec/autoexec.ipxe`. Confirm what `cwuri` actually resolves to when the binary is reached *via shim* rather than directly — the shim is loaded first, and whether `cwuri` follows the shim's path or iPXE's is worth observing rather than assuming.
2. **Whether the installer should place `autoexec.ipxe` at the TFTP root**, or whether the shim and `ipxe.efi` should live in their own subdirectory with a copy beside them. Falls out of (1).
3. **That MOK enrolment via `mmx64.efi` works from a PXE-booted shim**, not just from an installed distro.
[...]

Issue #957, opened 2026-08-01
Update to iPXE 2.0, stop overlaying its build system, and decide on Secure Boot


mastacontrola:
**Is your feature request related to a problem? Please describe.**

FOG builds iPXE from a fresh clone of upstream master and then pastes six of its own files over the top. One of those, `Makefile.housekeeping`, was a 2024 snapshot of upstream's build system, so every build was a 2024 build system driving whatever sources master happened to be on that day. That is the direct cause of #955, and it has quietly cost us upstream work: the last re-pin (`f61a90d97`, Jan 2026) reverted upstream's newer Secure Boot build mode — the one that *excludes known-insecure drivers* — back to the older scheme, purely as a side effect of copying an older file over a newer tree.

Separately, **iPXE 2.0.0 shipped on 6 March 2026** and it changes what is on the table for FOG:

- *"Support for UEFI Secure Boot via a dedicated iPXE shim"*, with signing automated as part of the release process and signed binaries published at [ipxe/shim](https://github.com/ipxe/shim/releases).
- Automatic `autoexec.ipxe` download.
- LoongArch64 and RISC-V support; a pile of new NIC drivers (Intel 100G, Marvell AQtion, Google GVE, Cadence GEM, DesignWare MAC).
- Reproducible builds.
- TLS v1.0 removed (breaking, but nothing FOG relies on).

**Describe the solution you'd like**

Three separable pieces. The first is done; the other two are decisions.

**1. Stop overlaying `Makefile.housekeeping`, and pass `NO_WERROR=1`.** (Done — see the linked commit.)

The overlay has never contained a FOG-specific line. Verified by building v2.0.0 on binutils 2.46 / gcc 16:

| build | result |
|---|---|
| v2.0.0 + FOG's overlaid `Makefile.housekeeping` | fails |
| v2.0.0 + upstream's own `Makefile.housekeeping` | fails, *same* error |
| v2.0.0 + upstream's, `NO_WERROR=1` | **builds** |
| v2.0.0 + upstream's, `NO_WERROR=1`, `CERT=`/`TRUST=` (the `-S/--force-https` path) | **builds** |

The failure is `-Werror=unused-but-set-variable` in `drivers/net/w89c840.c` — a compiler-vs-upstream problem in a driver nobody has touched in a decade, not something FOG should be carrying a patched build system for. `NO_WERROR` is upstream's own knob; FOG simply never passed it. That closes #955.

[...]

Issue #850, opened 2026-06-20
refactor(installer): variablize fogprogramdir through installer and derive PHP constants from it


mastacontrola:
## Background

The FOG base path `/opt/fog` is currently hardcoded in multiple places:

- `lib/common/config.sh` sets `fogprogramdir="/opt/fog"` but this doesn't flow into PHP
- `packages/web/lib/fog/system.class.php` defines `FOG_BASE_DIR` and `FOG_CACHE_DIR`
  as hardcoded strings (introduced in #849)
- Various `functions.sh` blocks reference `/opt/fog/service`, `/opt/fog/log`,
  `/opt/fog/cache` etc. as literal strings rather than using `$fogprogramdir`

## Problem

Non-standard installs (different base path) require manual edits across multiple files.
There is no single source of truth for the FOG base directory that both the shell installer
and the PHP runtime share.

## Proposed Solution

- Audit all hardcoded `/opt/fog` references in `lib/common/functions.sh` and replace with
  `$fogprogramdir` (already defined in `config.sh`)
- Have the installer write `fogprogramdir` into a PHP-readable location — either the
  generated `service/etc/config.php` or a new `packages/web/commons/fogpaths.php` — as a
  `define()` so the PHP runtime knows the real base path
- Update `system.class.php` to derive `FOG_BASE_DIR` from that generated define rather
  than hardcoding it, with `/opt/fog` as fallback if not defined
- All other path constants (`FOG_CACHE_DIR` etc.) derive from `FOG_BASE_DIR`

## Notes

- High blast radius — touches installer, service setup, symlink creation, log paths,
  tftp setup, and PHP boot chain. Needs careful testing across all supported distros.
- This is explicitly out of scope for #849 which hardcoded the path
  to keep that PR focused and reviewable.
- The install validation repo (`fogproject-install-validation`) should catch regressions
  if tests are run after changes here.

## Acceptance Criteria

- [ ] `fogprogramdir` change in `.fogsettings` is fully respected by installer
- [ ] PHP runtime derives `FOG_BASE_DIR` from installer-written value, not hardcoded string
- [ ] Fresh install and upgrade paths both work on all supported distros
- [ ] No hardcoded `/opt/fog` strings remain in `functions.sh`

mastacontrola:
Prerequisite landed: #849 is implemented on `working-1.6` in 86b26c2db.

The hardcoded constants this issue will variablize now exist in `packages/web/lib/fog/system.class.php::__construct()`:

[...]

PR #961, opened 2026-08-02
Keep FOS kernels signed for Secure Boot, and make enrolment a short visit

darksidemilk:
Follows the guide merged for #957 / #960. Companion changes on the same branch name in [FOGProject/fos](https://github.com/FOGProject/fos) (build-time signing) and [FOGProject/fog-docs](https://github.com/FOGProject/fog-docs) (the rewritten guide).

## The problem

A site running UEFI Secure Boot signs the FOS kernels with its own key and enrols that key per machine. Both halves were manual, and the signing half quietly came undone:

- `lib/common/functions.sh` copies freshly downloaded, **unsigned** kernels over `service/ipxe` on every install *and* every upgrade.
- The web **Kernel Update** page does the same over FTP.

Clients stopped booting the next time anyone updated FOG, with nothing on the server to explain it. The guide names this as the single most common way the setup breaks.

## What changed

**Signing survives upgrades.** `--secure-boot-key` / `--secure-boot-cert` are stored in `.fogsettings`, so both paths re-sign from then on. `_resignKernels()` skips images that already carry the configured signature — that is what stops a second run stacking a second signature — and refreshes its `.unsigned` snapshot each time, so an upgrade cannot re-sign the *previous* version over the new one.

**The web path signs without ever seeing the key.** The Kernel Update page runs as the web user, which must never be able to read the signing key. It signs through `/opt/fog/bin/fog-sign-kernel`, root-owned and reachable via a sudoers rule scoped to that one command. The helper **takes no arguments** — every path comes from a root-only config — so the caller cannot redirect it at a different key or file. The sudoers drop-in is validated with `visudo` before install, because a malformed one breaks `sudo` for the whole machine. Signing happens before the FTP upload, since the TFTP target may be a remote storage node.

> [!IMPORTANT]
> This protects the **key**, not the act of signing. Anyone who can already run code as the web server can have a kernel of their choosing signed. What they cannot do is walk off with the key and sign things elsewhere or later. The helper and the guide both say so plainly rather than implying more. Closing the remaining gap would mean verifying the staged image against a known-good hash before signing.

A signing failure **refuses the update** rather than installing a kernel clients cannot boot.

**Enrolment gets a kit.** The public certificate, a script and a `.desktop` launcher, published under `service/secureboot` and linked from a new **FOG Configuration → Secure Boot** page that displays the certificate fingerprint. The script prints the same fingerprint and makes the technician confirm it before trusting anything — that comparison is the whole security of the step, and the old guide only suggested eyeballing the CN in MokManager. It runs from a stock Ubuntu or Debian live USB, which already boots with Secure Boot **on**, so the visit no longer starts by turning Secure Boot off. Only the certificate is published; the private key never goes near the web root.

**`--secure-boot-stage`** optionally stages the signed chain: iPXE's Microsoft-signed shim, plus upstream's signed `snponly.efi` pulled from `ipxeboot.tar.gz`. Opt-in and never fatal — it fetches pinned third-party binaries, and an upstream hiccup must not fail an install.

The shim is installed as **`snponly-shimx64.efi`**, because the shim picks its second stage by stripping `-shim` from its own filename. That is upstream's own mechanism: `ipxeboot.tar.gz` ships `ipxe-shim.efi` and `snponly-shim.efi` as two symlinks to a single `shimx64.efi` for exactly this reason. Renaming does not disturb the signature, which covers file contents.

The shim comes from the `ipxe/shim` release rather than the tarball's bundled copy. Reading both PE certificate tables: the standalone build has two `WIN_CERTIFICATE` entries (Microsoft UEFI CA **2011 and 2023**), the bundled one has a single 2011 entry — no use on hardware shipping only the 2023 certificate.

> [!NOTE]
> Staging `snponly` rather than the all-drivers `ipxe.efi` means Secure Boot clients get **the same driver model FOG already serves everyone else** — it binds only the firmware's UEFI network protocol instead of replacing the NIC driver. These machines stop being a special case.

**The Secure Boot DHCP class is emitted commented out** in both the Kea sample and the ISC config. Option 93 carries the client architecture and nothing else, so there is no way to tell from a request whether Secure Boot is on — a site has to opt specific machines in once their firmware has the MOK enrolled.

## Verified

- `bash -n` on `installfog.sh` and `functions.sh`; `php -l` on both PHP files.
- `_resignKernels()` exercised in a sandbox with stubbed `sbsign`/`sbverify` across all three flows: fresh install signs once, a re-run with no new download skips, and an upgrade signs the **new** kernel rather than re-signing the stale `.unsigned`. An earlier revision double-signed on re-run; that is what the `sbverify` skip fixes.
- `fog-sign-kernel` guards tested: missing config, nothing staged, a symlink pointing outside the staging directory (refused, target untouched), and caller-supplied arguments (ignored).
- The staging sequence run against the real 12 MB `ipxeboot.tar.gz`: the `tar` path extracts, and both staged binaries carry the signatures described above.
- The generated Kea sample parses as valid JSON once `#` comments are stripped.
- The certificate fingerprint PHP computes matches `openssl x509 -fingerprint -sha256` exactly — the technician compares these two, so they have to agree.

## Not verified

No end-to-end run against a real FOG server or a Secure Boot client. The install → upgrade → Kernel Update cycle and the live USB enrolment both want testing on real hardware — see #960, which is already open for exactly that.

## Two corrections since this was opened

**A signed `snponly.efi` exists** (dc90e67). An earlier revision of this PR staged the all-drivers `ipxe.efi`, extracted from the release's `-sb.usb` disk image by parsing the MBR for the first partition's LBA start and loop-mounting at that offset. That was based on the belief that upstream published no signed `snponly` equivalent — the reason [ipxe/ipxe#1776](https://github.com/ipxe/ipxe/issues/1776) was filed, and why it is closed as not planned. The signed binaries ship in `ipxeboot.tar.gz` under `<arch>-sb/`, for arm64 as well as x86_64. The loop-mount machinery is deleted; a tarball needs none of it.

**Rebased onto `dev-branch`.** The branch was originally cut from `stable`, which predates 182701d (iPXE moved to the `fog-ipxe` repo) and 3c0b4ce (`autoexec.ipxe` hard-linked at the TFTP root). Both collided: the staging function no longer sources `src/ipxe/src/ipxescript`, which no longer exists, and its redundant root hard-link is gone — it now hard-links from `$tftpdirdst/autoexec/autoexec.ipxe`, matching the convention 3c0b4ce set.

https://claude.ai/code/session_01HZ2jZ7qgen6pXrbtmxShaw

### DIFF
--- a/bin/installfog.sh
+++ b/bin/installfog.sh
@@ -108,11 +108,15 @@ usage() {
     echo -e "\t-X    --exitFail\t\tDo not exit if item fails"
     echo -e "\t-T    --no-tftpbuild\t\tDo not rebuild the tftpd config file"
     echo -e "\t-F    --no-vhost\t\tDo not overwrite vhost file"
+    echo -e "\t      --secure-boot-key\t\tPrivate key used to re-sign the FOS"
+    echo -e "\t                       \t\t\tkernels for UEFI Secure Boot"
+    echo -e "\t      --secure-boot-cert\tCertificate matching --secure-boot-key"
+    echo -e "\t                        \t\t\t(both are required together)"
     exit 0
 }
 
 shortopts="h?odEUHSCKYyXxTPFf:c:W:D:B:s:e:b:N:"
-longopts="help,uninstall,ssl-path:,oldcopy,no-vhost,no-defaults,no-upgrade,no-htmldoc,force-https,recreate-keys,recreate-CA,recreate-Ca,recreate-cA,recreate-ca,autoaccept,file:,docroot:,webroot:,backuppath:,startrange:,endrange:,bootfile:,no-exportbuild,exitFail,no-tftpbuild"
+longopts="help,uninstall,ssl-path:,oldcopy,no-vhost,no-defaults,no-upgrade,no-htmldoc,force-https,recreate-keys,recreate-CA,recreate-Ca,recreate-cA,recreate-ca,autoaccept,file:,docroot:,webroot:,backuppath:,startrange:,endrange:,bootfile:,no-exportbuild,exitFail,no-tftpbuild,secure-boot-key:,secure-boot-cert:"
 
 optargs=$(getopt -o $shortopts -l $longopts -n "$0" -- "$@")
 [[ $? -ne 0 ]] && usage
@@ -270,6 +274,26 @@ while :; do
             novhost="y"
             shift
             ;;
+        --secure-boot-key)
+            if [[ -f $2 ]]; then
+                ssecureBootKey="$2"
+            else
+                echo "$1 requires a readable private key file after"
+                usage
+                exit 3
+            fi
+            shift 2
+            ;;
+        --secure-boot-cert)
+            if [[ -f $2 ]]; then
+                ssecureBootCert="$2"
+            else
+                echo "$1 requires a readable certificate file after"
+                usage
+                exit 3
+            fi
+            shift 2
+            ;;
         --)
             shift 
             break 
@@ -400,6 +424,26 @@ esac
 [[ -n $sbackupPath ]] && backupPath=$sbackupPath
 [[ -n $sexitFail ]] && exitFail=$sexitFail
 [[ -n $snoTftpBuild ]] && noTftpBuild=$snoTftpBuild
+[[ -n $ssecureBootKey ]] && secureBootKey=$ssecureBootKey
+[[ -n $ssecureBootCert ]] && secureBootCert=$ssecureBootCert
+
+# Secure Boot signing is opt-in and only meaningful as a pair. Refuse half a
+# pair rather than silently leaving kernels unsigned on a server whose admin
+# believes they are signed -- that failure only shows up at a client, as a
+# Security Policy Violation with nothing on the server to explain it.
+if [[ -n $secureBootKey || -n $secureBootCert ]]; then
+    if [[ -z $secureBootKey || -z $secureBootCert ]]; then
+        echo " * --secure-boot-key and --secure-boot-cert must be set together"
+        exit 9
+    fi
+    for sbfile in "$secureBootKey" "$secureBootCert"; do
+        if [[ ! -r $sbfile ]]; then
+            echo " * Cannot read Secure Boot signing file: $sbfile"
+            exit 9
+        fi
+    done
+    unset sbfile
+fi
 
 [[ -f $fogpriorconfig ]] && grep -l webroot $fogpriorconfig >>$error_log 2>&1
 case $? in
--- a/lib/common/functions.sh
+++ b/lib/common/functions.sh
@@ -3274,6 +3274,18 @@ _installSecureBootSigner() {
         echo "Failed"
         return 0
     }
+    # Point the helper at this install's config. It takes no arguments on
+    # purpose -- that is what stops a compromised web server naming its own key
+    # -- so the path has to be baked in here rather than passed at call time.
+    # Quoted: $fogprogramdir may contain a space, and `CONF=/a/fog custom/x`
+    # assigns only "/a/fog" and then tries to RUN "custom/x". bash -n does not
+    # catch that -- it is valid syntax, just not what anyone meant.
+    sed -i "s|^CONF=.*|CONF=\"${conf}\"|" "$helper" >>$error_log 2>&1
+    if ! grep -qxF "CONF=\"${conf}\"" "$helper"; then
+        echo "Failed"
+        echo " * Could not set the config path in $helper."
+        return 0
+    fi
     # Root-owned, root-readable only: the web user learns nothing about where
     # the key lives, and cannot rewrite these paths to point somewhere else.
     # SECUREBOOT_CERT is the normalised PEM -- sbsign cannot read DER.
--- a/lib/common/functions.sh
+++ b/lib/common/functions.sh
@@ -3194,6 +3194,39 @@ _publishSecureBootKit() {
     chown -R "${apacheuser}":"${apacheuser}" "$kitdir" >>$error_log 2>&1
     echo "Done"
 }
+# Normalise --secure-boot-cert to PEM and echo the path.
+#
+# sbsign and sbverify read certificates with PEM_read_bio_X509 and reject DER
+# outright:
+#
+#   $ sbsign --key MOK.priv --cert MOK.der --output out.efi in.efi
+#   Can't load certificate from file 'MOK.der'
+#   error:0480006C:PEM routines:get_name:no start line ... Expecting: CERTIFICATE
+#
+# mokutil and MokManager want the opposite -- DER. So an admin following any
+# Secure Boot guide ends up holding one of each, with nothing telling them which
+# tool takes which, and handing the wrong one to --secure-boot-cert fails at
+# signing time with an OpenSSL error that says nothing about the format.
+#
+# Convert once here rather than pushing the distinction onto the user: the flag
+# accepts either, _resignKernels and the signing helper get PEM, and
+# _publishSecureBootKit still converts to DER for enrolment.
+_secureBootCertPem() {
+    local pem="${fogprogramdir}/.fog-secureboot.pem"
+    [[ -z $secureBootCert ]] && return 1
+    mkdir -p "$fogprogramdir" >>$error_log 2>&1
+    if openssl x509 -in "$secureBootCert" -inform pem -noout >/dev/null 2>&1; then
+        cp -f "$secureBootCert" "$pem" >>$error_log 2>&1 || return 1
+    elif ! openssl x509 -in "$secureBootCert" -inform der -outform pem \
+            -out "$pem" >>$error_log 2>&1; then
+        return 1
+    fi
+    # World-readable on purpose: a certificate is public, and it is the private
+    # key next to it that the 0600 config protects.
+    chown root:root "$pem" >>$error_log 2>&1
+    chmod 0644 "$pem" >>$error_log 2>&1
+    echo "$pem"
+}
 # Re-sign the FOS kernels for UEFI Secure Boot.
 #
 # The kernels above are downloaded unsigned on every install AND every upgrade,
@@ -3214,11 +3247,16 @@ _publishSecureBootKit() {
 # Removes the helper, its config and the sudoers rule again when signing is
 # turned off, so disabling the feature actually disables the privilege.
 _installSecureBootSigner() {
-    local bindir="/opt/fog/bin"
+    # $fogprogramdir, not a "/opt/fog" literal: GH-850 made the base path
+    # installer-driven, and hardcoding it here would scatter the helper, its
+    # config and the staging directory back into /opt/fog on a server installed
+    # anywhere else.
+    local bindir="${fogprogramdir}/bin"
     local helper="${bindir}/fog-sign-kernel"
-    local conf="/opt/fog/.fog-secureboot"
-    local stagedir="/opt/fog/secureboot-staging"
+    local conf="${fogprogramdir}/.fog-secureboot"
+    local stagedir="${fogprogramdir}/secureboot-staging"
     local sudoersfile="/etc/sudoers.d/fog-secureboot"
+    local certpem
 
     if [[ -z $secureBootKey || -z $secureBootCert ]]; then
         rm -f "$helper" "$conf" "$sudoersfile" >>$error_log 2>&1
@@ -3226,6 +3264,11 @@ _installSecureBootSigner() {
     fi
 
     dots "Installing Secure Boot signing helper"
+    certpem=$(_secureBootCertPem) || {
+        echo "Failed"
+        echo " * Could not read $secureBootCert as a certificate (PEM or DER)."
+        return 0
+    }
     mkdir -p "$bindir" >>$error_log 2>&1
     install -o root -g root -m 0755 ../packages/secureboot/fog-sign-kernel "$helper" >>$error_log 2>&1 || {
         echo "Failed"
@@ -3233,9 +3276,10 @@ _installSecureBootSigner() {
     }
     # Root-owned, root-readable only: the web user learns nothing about where
     # the key lives, and cannot rewrite these paths to point somewhere else.
+    # SECUREBOOT_CERT is the normalised PEM -- sbsign cannot read DER.
     {
         echo "SECUREBOOT_KEY=${secureBootKey}"
-        echo "SECUREBOOT_CERT=${secureBootCert}"
+        echo "SECUREBOOT_CERT=${certpem}"
         echo "SECUREBOOT_STAGING=${stagedir}"
     } > "$conf"
     chown root:root "$conf" >>$error_log 2>&1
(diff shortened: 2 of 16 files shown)


darksidemilk:
Pushed dc90e67, which changes what this stages. The PR description above is now out of date in one respect, so flagging it rather than letting a reviewer hit it.

**A signed `snponly.efi` does exist.** It ships in `ipxeboot.tar.gz` under `x86_64-sb/` (and `arm64-sb/`), signed by the iPXE Secure Boot Intermediate G1A. The original code looked in the release's `-sb.usb` disk image, which carries only `ipxe.efi` — hence the wrong conclusion, and hence [ipxe/ipxe#1776](https://github.com/ipxe/ipxe/issues/1776) being closed as not planned.

What changed:

- Stages `x86_64-sb/snponly.efi` from `ipxeboot.tar.gz` instead of extracting `IPXE.EFI` from the disk image. **The MBR-offset parsing and loop-mount are gone** — a tarball needs neither, so that whole block is deleted.
- Installs the shim as `snponly-shimx64.efi`. The shim picks its second stage by stripping `-shim` from its own filename; upstream ships `ipxe-shim.efi` and `snponly-shim.efi` as two symlinks to one `shimx64.efi` for exactly this reason. Renaming does not disturb the signature.
- DHCP boot file in both the Kea sample and the ISC config is now `secureboot/snponly-shimx64.efi`.
- The comments claiming no signed `snponly` exists are corrected.

Still pulling the shim from the `ipxe/shim` release rather than the tarball's bundled copy. Reading both PE certificate tables: the standalone build has two `WIN_CERTIFICATE` entries (MS UEFI CA 2011 **and** 2023), the bundled one has a single 2011 entry — no use on hardware shipping only the 2023 certificate.

The practical upshot is that Secure Boot clients now get the same driver model as everyone else. `snponly` binds only the firmware's UEFI network protocol rather than replacing the NIC driver, which is what FOG already serves by default, so these machines stop being a special case — and the takeover hang that motivated #1776 stops being a concern.

Verified the exact staging sequence against the real 12 MB tarball: the `tar` path extracts, and both staged binaries carry the signatures above.

---

mastacontrola:
Reviewed this against a full end-to-end run on the lab — firmware → shim → signed `snponly.efi` → `autoexec.ipxe` → `boot.php` → MOK-signed `bzImage` → FOS → 42 GB deployed, Secure Boot enforcing throughout (#960, now closed). The approach here is right and the mechanism works exactly as this PR describes. Six things from doing it by hand.

## 1. `--secure-boot-cert` cannot be a DER file, and the guide says it should be

This is the blocking one, and it spans all three PRs.

`sbsign` and `sbverify` read certificates with `PEM_read_bio_X509`. They cannot parse DER:

```
$ sbsign --key MOK.priv --cert MOK.der --output out.efi in.efi
Can't load certificate from file 'MOK.der'
error:0480006C:PEM routines:get_name:no start line:crypto/pem/pem_lib.c:792:Expecting: CERTIFICATE

$ sbverify --cert MOK.der out.efi
Can't load certificate from file 'MOK.der'
```

DER is what `mokutil` and MokManager want. PEM is what `sbsign` wants. The guide's Step 1 produces only `MOK.der` (`-outform DER`), and then:

- fog-docs#84 documents `--secure-boot-cert /root/fog-secureboot/MOK.der` and `--sign-cert /root/fog-secureboot/MOK.der`
- this PR passes `$secureBootCert` straight to `sbverify` (line ~328) and `sbsign` (line ~335), and `fog-sign-kernel` does the same (line ~627)
- FOGProject/fos#130 does `sbsign --key "$signKey" --cert "$signCert"` (line ~83)

So an admin who follows the guide exactly gets a signing failure on first use. Worth noting the failure is *loud* in this PR — a signing failure refuses the update, which is the right call and means this is annoying rather than dangerous.

This PR already does the right normalisation for the enrolment kit:

```bash
if openssl x509 -in "$secureBootCert" -inform der -noout >/dev/null 2>&1; then
    cp -f "$secureBootCert" "${kitdir}/MOK.der"
elif openssl x509 -in "$secureBootCert" -outform der -out "${kitdir}/MOK.der"; then
```

It needs the same thing in the other direction — a canonical PEM derived once at config time and handed to `sbsign`/`sbverify` — so that `--secure-boot-cert` accepts either format and the admin never has to care. Same for `FOS_SIGN_CERT` in fos#130.

## 2. `restorecon` after staging

`_stageSecureBootChain()` writes into `/tftpboot` with `curl` and `cp`, which yields `default_t` rather than `tftpdir_t`. On an enforcing host `in.tftpd` is then denied read and the client sees a bare "file not found" with nothing server-side to explain it.

This is not hypothetical — the audit log caught it during the #960 run:

```
avc: denied { read } for comm="in.tftpd" name="default.ipxe"
  tcontext=unconfined_u:object_r:default_t:s0 permissive=1
```

Separately and more broadly: **91 of 96 files under `/tftpboot` are already `default_t`** on a stock install, and `checkSELinux()` only offers to set SELinux permissive. That is its own bug and I have not filed it yet, but this PR shouldn't add to the pile.

## 3. Only create a `-shim` alias whose stage 2 was actually staged

Installing the shim as `snponly-shimx64.efi` is correct, and I verified the naming works: `automatic_next_path()` matches `-shim` plus up to four more characters before `.efi` (`dot - hyphen <= 9`), so both `snponly-shim.efi` and `snponly-shimx64.efi` resolve to `snponly.efi`.

If anyone later adds `ipxe-shim.efi` for symmetry, it must be conditional on `ipxe.efi` having been staged — otherwise it produces a chain that TFTP-404s at stage 2.

One limitation worth documenting: `automatic_next_path()` is called **only** from `netboot.c` and `httpboot.c`. There is no local-filesystem path, so a shim booted from a USB stick or ESP ignores the rename entirely and goes to its compiled-in default of `ipxe.efi`. That matters for FOGProject/fos#129 — a Secure Boot USB built this way silently gets the all-drivers binary.

## 4. Take the binaries from a fog-ipxe release asset, not from upstream at install time

Maintainer direction, and I agree with it: rather than each install fetching from `ipxe/shim` and `ipxe/ipxe`, the fog-ipxe release workflow should pull the pinned upstream artifacts once and publish them as a release asset set (e.g. a separate `fog-ipxe-secureboot.tar.gz`).

The installer already downloads a fog-ipxe release, so this costs zero extra network calls, inherits `downloadfiles()`'s sha256 + retry, and makes offline installs work by pre-placing a file. It also weakens the argument for keeping staging opt-in, since *placing files* is inert until DHCP points at them.

For that workflow to be safe rather than just relocating the risk:

- **fail on sha256 mismatch** against a recorded expected value, so an upstream asset silently republished under the same tag is caught at release rather than propagated
- **assert the signer** by parsing the PE certificate table — `Microsoft Corporation UEFI CA 2011`/`2023` for the shim, `iPXE Secure Boot Automatic Code Signing G1A` for `snponly.efi`. Without this, an upstream test-signed build propagates silently and surfaces at a client as `Security Policy Violation` with nothing to explain it
- **rename at release time**, so the filename-is-the-mechanism trap lives in one reviewed place instead of the installer
- **emit a manifest** (upstream URL, sha256, signer, tag). Rehosting a Microsoft-signed binary is normal — every distro does it — but it needs to be auditable, and the manifest is also the SBAT revocation trail: when a shim generation is revoked you can tell which one a given release shipped

## 5. `--secure-boot-stage` is x86_64 only

`ipxeboot.tar.gz` ships the same file set under `arm64-sb/`, including `snponly.efi` and `snponly-shim.efi`.

## 6. The DHCP half is where the documented path earns its keep

Which filename gets served is a DHCP decision — the rewrite runs off `BootpBootFile`, not the filesystem. Emitting the Secure Boot class commented out is right (option 93 carries architecture only, so nothing in the request says whether Secure Boot is on), but it does mean all the staging is inert until someone edits DHCP by hand.

That is worth being loud about, because it is exactly what cost me two boot cycles on the lab: a per-MAC `dhcp-boot` override in a `zz-*.conf` lost to `ltsp.conf`, since dhcp-boot is first-tag-match-wins in file-read order **and** `pxe-service` overrides `dhcp-boot` entirely in proxyDHCP mode. The client silently got FOG's unsigned `snponly.efi`, Secure Boot refused it, and the only symptom was the boot manager returning with no message.

Also worth recording for the docs: this firmware reports arch **`00007` (`BC_EFI`)**, not `X86-64_EFI`.

---

Nothing above blocks the design, and #1 is the only thing I would call a must-fix before merge.

mastacontrola:
⚠️ **This PR now conflicts with shipped code and should not be merged as-is.** Three of its pieces were solved a different way while it sat open.

Landed on both branches today (working-1.6 `06a460275`, dev-branch `2121873d9`, plus FOGProject/fog-ipxe#1 and release `v2.0.0-fog.2`):

**1. `--secure-boot-stage` is superseded.** This PR adds `stageSecureBootBinaries()`, which curls `ipxe-shimx64.efi` straight from `ipxe/shim` at install time, behind an opt-in flag. That is now `downloadipxesecureboot()`, which takes the same binaries from the `fog-ipxe` release the installer already downloads.

The difference matters:

| | this PR | shipped |
|---|---|---|
| when | install time, every site, unsupervised | release time, once, supervised |
| source | two third-party release URLs | one asset the installer already fetches |
| integrity | none | sha256 + PE signer assertion, checked in CI on every PR |
| offline | impossible | works from a pre-placed tarball |
| opt-in | `--secure-boot-stage` | unconditional |

A fetch failure at install time surfaces weeks later as a machine that will not PXE boot. At release time it fails in front of a human. The signer assertion is the part I would not want to give up — an unsigned or test-signed upstream build otherwise propagates to every install and shows at the client as `Security Policy Violation` with nothing server-side to explain it.

**2. The `autoexec.ipxe` hard link is superseded**, and was incomplete. This PR links it into `secureboot/`. There are six directories needing a copy, not two — iPXE resolves the bare name against whatever directory the running `.efi` came from, so `autoexec/i386-efi/` and `autoexec/arm64-efi/` were being missed as well. All six are now one inode.

**3. The `restorecon` line is superseded by #963.** This PR has:

```bash
command -v restorecon >/dev/null 2>&1 && restorecon -R "$tftpdirdst"
```

That is the right instinct but it silently does nothing in the case that matters. `restorecon` exits 0 when it has no rule to apply, so a relocated `$tftpdirdst` — which has no fcontext rule — comes out still unreadable and still reporting success. `setSELinuxContext()` registers a rule via `semanage` first when policy has no mapping, uses `-RF`, and verifies the resulting label instead of trusting the exit code.

## What this PR should keep

Everything that is actually about signing, which is the part nothing else covers:

- `packages/secureboot/fog-sign-kernel` and the sudoers/root-key custody around it
- `--secure-boot-key` / `--secure-boot-cert`
- re-signing on web Kernel Update (`fogpage.class.php`, `fogconfigurationpage.class.php`)
- the MOK enrolment kit (`fog-enroll-mok.sh`, `.desktop`)
- the commented DHCP class — the path it references, `secureboot/snponly-shimx64.efi`, is unchanged and now exists by default, so the comment can drop "Requires `--secure-boot-stage`"

## Also still open on this PR

The DER bug from [my earlier comment](https://github.com/FOGProject/fogproject/issues/961#issuecomment-5157733444): `--secure-boot-cert` is passed straight to `sbsign`/`sbverify`, which read PEM only and reject DER outright. The docs hand users a `MOK.der`. Either normalise internally with `openssl x509 -inform DER -outform PEM` or reject DER with a message that says which format is wanted.

Rebase needed either way — `lib/common/functions.sh` has moved underneath this branch.

mastacontrola:
Reworked and force-pushed (`f62a592ec`), rebased onto current `dev-branch`.

## Removed

- **`_stageSecureBootChain()` and `--secure-boot-stage`.** Superseded by `downloadipxesecureboot()`.
- **The `autoexec.ipxe` hard link into `secureboot/`.** Superseded, and it was incomplete — six directories need a copy, not two. `configureTFTPandPXE` links all six now.
- **`restorecon -R "$tftpdirdst"`.** Superseded by #963's `setSELinuxContext()`.

The second commit on this branch (*"Stage the signed snponly.efi rather than the all-drivers build"*) was entirely about the removed function, so it dropped out in the rebase. The DHCP samples keep their commented Secure Boot class — the path is unchanged and now exists on every install, so only the "Requires `--secure-boot-stage`" caveat went.

## Fixed while here

**`--secure-boot-cert` now accepts DER as well as PEM.** This was the blocking bug from [my earlier comment](https://github.com/FOGProject/fogproject/issues/961#issuecomment-5157733444) and it is worth being explicit about, because the failure is so unhelpful:

```
$ sbsign --key MOK.priv --cert MOK.der --output out.efi test.efi
Can't load certificate from file 'MOK.der'
```

`sbsign`/`sbverify` read PEM only; `mokutil` and MokManager want DER. Every Secure Boot guide leaves an admin holding one of each with nothing saying which tool takes which. `_secureBootCertPem()` normalises once — signing paths get PEM, `_publishSecureBootKit` still emits DER for enrolment.

Verified both directions on a freshly generated key pair:

```
MOK.der -> Signature verification OK
MOK.pem -> Signature verification OK
```

and a non-certificate file is rejected with a message naming the file.

**`_installSecureBootSigner()` hardcoded `/opt/fog`** for the helper, its config and the staging directory. GH-850 made the base path installer-driven, so on a server installed anywhere else those three would have scattered back into `/opt/fog` while the rest of FOG lived elsewhere. Now `$fogprogramdir`.

## What this PR is now

Purely the signing half, which is the part nothing else covers:

- `fog-sign-kernel` + sudoers rule + root-only key custody
- `--secure-boot-key` / `--secure-boot-cert`, either certificate format
- re-signing on install *and* upgrade, and on web Kernel Update
- the MOK enrolment kit
- commented DHCP classes for ISC and Kea

The known limit stated in #962 is unchanged: the sudo helper protects the **key**, not the act of signing. Anyone who can run code as the web server can have a kernel of their choosing signed; they cannot walk off with the key.

Still untested end to end against an enforcing/Secure Boot server from a clean install — the chain itself is proven in #960, but that used hand-signed kernels rather than this automation.

mastacontrola:
Good catch — I fixed the shell half earlier and missed the PHP. Pushed `d36f94637`.

GH-850 made the base path installer-driven, but this feature had `/opt/fog` literals on **both** sides of the web/installer boundary. On a server installed anywhere else the two halves pointed at different directories, and the failure was silent in the worst way: `secureBootStagingDir()` returned `''` because the helper was not where it looked, so Kernel Update quietly shipped an **unsigned** kernel and Secure Boot clients stopped booting with nothing on the server to explain it.

Three places:

- **`fogpage.class.php`** derived the helper and staging paths from two `/opt/fog` literals → now `FOG_BASE_DIR`.
- **The `exec()`** invoking the helper had the path hardcoded too → now built from `FOG_BASE_DIR` and passed through `escapeshellarg()`. It is no longer a literal, `exec()` hands the string to a shell, and a base path with a space would otherwise split into two arguments — which the sudoers rule, matching the exact command, would then refuse.
- **`fog-sign-kernel`** read its config from `/opt/fog/.fog-secureboot` while the installer had already been changed to write it under `$fogprogramdir`. That mismatch was mine, introduced when I moved the installer side. The installer now rewrites the `CONF` line on install and verifies the rewrite landed.

Substituted rather than passed as an argument, because taking no arguments is that script's security property — it is what stops a compromised web server naming a key of its own. An environment variable would be no better; it puts the path back under the caller's control.

## One thing worth flagging

The substituted value is **quoted**, and it has to be. Unquoted, a base path with a space produces:

```bash
CONF=/a/fog custom/.fog-secureboot
```

which assigns only `/a/fog` and then tries to **run** `custom/.fog-secureboot`. That is valid shell, so `bash -n` passes it clean — I only found it by actually running the thing:

```
$ bash -c 'CONF=/a/fog custom/.conf; echo "CONF=[$CONF]"'
bash: line 1: custom/.conf: No such file or directory
CONF=[]
```

## Verified

End to end against a base directory whose path contains a space: installer `sed` lands and is confirmed, helper resolves its config, and it signs a kernel that `sbverify` then reports as `Signature verification OK`.

`fogconfigurationpage.class.php` was already correct — it uses `BASEPATH`, which is the web root and the right constant there. The enrolment kit has no base-dir references; it runs on the client. Pre-existing `/opt/fog` literals elsewhere in `functions.sh` are left alone — not part of this feature.

PR #1014, opened 2026-08-07
Update workflow: vhost/TFTP preservation, custom branches, and Let's Encrypt scaffolding

darksidemilk:
## Summary

Follow-ups to `bin/updatefog.sh`/`fog_git_path`/`fog_update_channel` (#1005/#1011), covering #1012 and #1013.

**#1012 — vhost/TFTP customization preservation, custom branches, sticky channel**
- Fixed a real bug: the nginx vhost write and Debian's `/etc/default/tftpd-hpa` write called `diffconfig()` without ever taking the `mv -fv` backup it needs to detect a change, so the existing "Changed configurations" warning (already correct for Apache's vhost and the cron reporting file) silently never fired for either.
- `updatefog.sh` now defaults to `-F/--no-vhost` when re-invoking `installfog.sh` (an update always has a pre-existing vhost); `--overwrite-vhost` opts back in.
- Added `--branch <name>` to check out an arbitrary branch for testing, independent of the tracked channel.
- `--channel` now actually persists (previously only affected the single run).
- Renamed channel values `stable/dev/beta` → `stable/staging/dev` to match the README's own Channel table instead of colliding with it.

**#1013 — cert/Let's Encrypt investigation and scaffolding**
- Corrected `docs/EXTERNAL_CA_AND_LETSENCRYPT.md`: iPXE's netboot HTTPS fetches are not coupled to FOG's own CA the way fog-client is — upstream iPXE's `CROSSCERT` fallback already validates a real Let's Encrypt certificate with no FOG-side change, independent of Secure Boot status. Also clarified FOG's SSL CA and the Secure Boot signing key are separate, independently-generated keys.
- Added `--hostname <name>` (non-interactive override of the vhost/cert hostname, previously only settable via an interactive prompt).
- Added repeatable `--extra-server-name <name>` for additive extra vhost/cert names, mirrored into `globalSettings` for GUI visibility.
- Added `bin/setupacme.sh`: bootstraps `acme.sh` to issue/renew the web vhost's **leaf** certificate against a CA already imported via `--external-ca`, never touching the CA itself, so fog-client's pinning survives renewals. Includes a persisted `acmeLeaf` marker so a later plain install/update doesn't silently regenerate and corrupt the ACME-managed leaf (except under `--recreate-keys`/`--recreate-ca`, where regenerating a fresh self-signed pair is the correct fallback).

## Test plan

No CI/test framework exists for these shell scripts beyond `fogproject-install-validation`'s distro matrix. Verification performed:
[...]

Issue #978, opened 2026-08-03
UEFI PXE boot fails after upgrading to 1.5.10.2149 — HTTP-to-HTTPS redirect causes “Permission denied” with older iPXE binary


RogerioKoglin:
# UEFI PXE boot fails after upgrading to 1.5.10.2149 — HTTP-to-HTTPS redirect causes "Permission denied" with older iPXE binary

## Summary

After upgrading FOG from `1.5.10.1903` to `1.5.10.2149`, UEFI PXE boot stopped working on at least one client.

The binaries generated by the new installer did not boot correctly on the affected machine. After restoring the previous `/tftpboot` binaries, the client successfully loaded `ipxe.efi`, but iPXE failed when chaining to FOG with:

```text
Permission denied
```

The issue was resolved only after restoring an Apache rewrite exception that allows `/fog/service/` endpoints to remain accessible over HTTP.

I would like to understand whether this is:

1. A bug or regression in FOG `1.5.10.2149`;
2. An incorrect DHCP/PXE configuration on my side;
3. A compatibility limitation in the affected clients' UEFI firmware;
4. Or an unsupported configuration because HTTPS is now expected for the entire iPXE flow.

## Environment

- FOG server type: Normal/Master server
- Previous FOG version: `1.5.10.1903`
- Current FOG version: `1.5.10.2149`
- Previous database schema: `274`
- Current database schema: `277`
- Operating system: Debian GNU/Linux 13 (Trixie)
- Kernel: `6.12.100+deb13-amd64`
- Apache: `2.4.68`
- PHP: `8.4`, using PHP-FPM
- MariaDB: `11.8.6`
- TFTP server: `tftpd-hpa`
- DHCP server: External; FOG does not provide DHCP
- FOG server address: `<FOG_SERVER_IP>`
- FOG hostname: `<FOG_HOSTNAME>`
- Network interface: `bond0`, 802.3ad/LACP with four 1 Gbps interfaces
- Client boot mode: UEFI
- DHCP boot file used for the affected client: `ipxe.efi`

The external DHCP server provides the FOG server address as the next server/option 66 and `ipxe.efi` as the UEFI boot filename/option 67.

## Upgrade procedure

The server operating system was fully upgraded before updating FOG.

FOG was then upgraded from a clean checkout of the official `stable` branch:

```text
Version: 1.5.10.2149
Schema: 277
Commit: df82a454981637c82ce84c869ee676947ba7c44a
```

[...]

PR #1015, opened 2026-08-08
Three-zone PKI (Web TLS / Client Communication / Secure Boot) + install-time customization preservation

darksidemilk:
Splits FOG's certificate handling into three independent zones (Web TLS,
Client Communication, Secure Boot) and moves customization preservation into
`installfog.sh` itself. Stacked: this branch includes
`customization-preservation`, because the PKI work depends on its vhost
managed-block.

Continuation notes: `docs/superpowers/CONTEXT-1013-pki-and-customizations.md`.

## The finding that drove this

`.srvprivate.key` is **both** the web vhost's TLS private key and the key
`FOGBase::certDecrypt()` opens on every fog-client `authorize()` handshake.
Confirmed on a live server by modulus comparison — it pairs with
`srvpublic.crt`, not `ca.cert.pem`.

So today, overwriting that file — an ACME renewal with `--key-file`,
`--recreate-keys`, a purchased cert dropped in place — breaks client
authentication while installing a perfectly valid certificate, with nothing in
the logs connecting the two. Pointing the vhost at a *different* file is safe;
overwriting FOG's is not. This is a live bug independent of the redesign.

## PKI

```
FOG Server ROOT CA
├── FOG Web CA          → web server certificate (vhost)
├── FOG Server CA       → ca.cert.der (pinned by fog-client)
│   └── FOG Client Communication → the key certDecrypt() opens
└── FOG Secure Boot CA  → MOK.der (enrolled in firmware)
    └── FOG Project Secure Boot Signing → signs kernels, --addcert'd
```

Default on **fresh** installs. An **existing** server always resolves to
`flat` and is never switched automatically — switching one underneath a fleet
that pinned the old CA is the one thing this must not decide on its own.
`--split-pki` opts in, `--legacy-pki` opts a fresh install out; both fully
supported. Per-zone `--web-ca-*`/`--client-ca-*`/`--root-ca-*` let an admin
replace any single zone.

Secure Boot now enrols the **issuer** rather than the signer, so signing
leaves can be rotated, revoked, or issued per storage node without a firmware
trip. Landing it before Secure Boot ships in a stable release is deliberate:
with no release out there is no enrolled fleet, so the restructure is free
now and costs a MokManager visit to every machine later.

Also removes `bin/setupacme.sh` — FOG should read the certificate it is given,
not run an ACME client.

## Customization preservation

Backup/restore moved into `installfog.sh`, so a bare `./installfog.sh`
upgrade is protected too — previously only `updatefog.sh` did any of this.

- Vhost written as a **FOG-managed block**; anything outside the markers
  survives, so admins keep security fixes *and* their own directives. Ends the
  all-or-nothing `-F` trade-off.
- iPXE background keyed to the real `FOG_IPXE_BG_FILE` value, not `"bg.png"`.
- Kernels: previous release kept as `bzImage.<release>` beside the new one,
  plus bounded generations and `bin/restorekernel.sh`.
- Kernel/init fields in the web UI are now **dropdowns** of what is on disk.
  Blank still means "inherit the default" and is never pre-selected.
- Optional `custom.ipxe` hook so PXE customization does not mean editing a
  generated file.
- Admin-supplied Secure Boot keys copied out of `$webdirdest` before the
  installer deletes it.

## Verified on a real server

CentOS Stream 9 / Apache, across repeated installs and a full
uninstall→purge→fresh cycle: all chains verify; comm key and vhost key are
provably different keypairs; a signed kernel carries both the leaf and the
intermediate (`sbverify --list`); `--legacy-pki` reproduces the original flat
CA exactly; an existing install's CA is byte-identical after upgrade; custom
vhost block, renamed background and custom kernels all survive; FOG's own
shipped files are never reverted.

**Eleven bugs were found this way and none by sandbox testing** — three would
have shipped as silent data loss. The recurring cause was harnesses that
reproduced a function without reproducing the sequence that calls it.

## Verified on real UEFI hardware

Secure Boot works via **both** enrolment routes with only the **intermediate**
trusted — enrolled as `MOK.der` through MokManager, or written into `db`
through the Setup Mode PK/KEK/db path. Machines boot FOG's leaf-signed
kernels.

That settles the question this restructure rested on and that no local testing
could answer: firmware and shim accept a chain terminating at the enrolled CA
rather than demanding the exact signer. A signing leaf can therefore be
rotated, revoked, or issued per storage node with no firmware trip — which is
the entire reason for enrolling the issuer instead of the signer. It also
exercises PXE boot by implication.

## fog-client confirmed, with one follow-up for `zazzles`

fog-client authenticates correctly against a split server — it fetches the
comm certificate from the path it always has, so the separation needed **no
client-side change**.

At install it does add `FOG Server CA` — the Client Communication
*intermediate* — to the Windows Root store, rather than `FOG Server ROOT CA`.
Nothing is broken, but it is the wrong anchor and costs two things:

- **Rotation.** Trusting an intermediate as an anchor means replacing it
  requires re-pushing trust to every client — exactly the cost the Secure Boot
  zone just removed by enrolling the issuer.
- **HTTPS by default.** The client trusts only the Client zone, which does not
  sign the web certificate (the **Web CA** does), so the web certificate is
  untrusted. If the client trusted the **root**, every zone beneath it would
  validate and an all-FOG-PKI install could enable HTTPS out of the box.
  **Confirmed** by adding the root to the Windows trust store by hand -- HTTPS
  then works, so the mechanics are proven and only the client-side choice of
  certificate remains.

That change belongs in `zazzles`, not this PR. Nothing here needs to change
for it: the root is already published in the chain, and `ca.cert.der` keeps
carrying the intermediate for existing pinning.

## Still not verified

**nginx.** All vhost work was verified on Apache only; the managed-block
splice and the `netbootproto` redirect exclusion both have nginx branches that
have never executed.

https://claude.ai/code/session_01YE3jeTexYEcAxWFbZcZgpZ


mastacontrola:
I picked up the nginx verification you flagged as never executed, and it turned up two real ones. I've pushed both fixes to the branch rather than leave you a list — shout if you'd rather I'd left them as comments. Reasoning below so you can check my working.

### 1. The netboot exclusion never fired — `6ffaba3b9`

The `location ^~ ${webroot}service/ipxe/` block was emitted ahead of the catch-all on the basis that first-emitted wins, but the catch-all is a server-level `return 308`. nginx runs a server-level `return` in the **server rewrite phase**, which is before location selection — so emission order can't buy anything, the return fires for every request, and the location block is unreachable.

That produced exactly the failure the exclusion exists to prevent: iPXE fetching `boot.php` over HTTP got 308'd to the HTTPS it can't validate. Same shape as #978.

Measured against real nginx rather than reasoned about:

| request | before | after |
|---|---|---|
| `/fog/service/ipxe/boot.php` | **308 → https** | 200, handled |
| `/fog/management/index.php` | 308 → https | 308 → https |
| `/` | 308 → https | 308 → https |

Fix is to scope the redirect as `location / { return 308 ...; }` — `^~` on the iPXE prefix then outranks `/`, so location matching actually runs and the exclusion wins. I also confirmed the nested `location ~ \.php$` that `include ${phploc}` splices into the iPXE block parses clean under `nginx -t`.

Your Apache branch is unaffected — `RewriteCond` genuinely does guard the next `RewriteRule`. That asymmetry is almost certainly why an Apache-only verification pass couldn't have caught this one.

### 2. `spliceManagedBlock` silently no-ops on a CRLF vhost — `48f28b648`

The "does this file have markers" test is `grep -F`, which is a substring match and so tolerates a trailing CR or a stray space. The action taken on that decision is awk comparing whole lines. Two matchers with different ideas of what the marker line is.

A vhost saved with CRLF endings passes the grep, matches nothing in the awk, and falls through every rule: file copied byte-for-byte, freshly generated vhost silently discarded, return 0, installer prints success. Trailing whitespace after a marker does the same thing.

What makes it worth fixing rather than filing: `SUPPORTED_CUSTOMIZATIONS.md` invites admins into this exact file. One edit from a Windows box, or one backup restored through one, and every subsequent install or update quietly stops updating the vhost — stranding whatever the managed block is carrying, including the `maintenance/` deny rules.

The fix normalizes trailing CR/whitespace for the *comparison* only; `$0` is still what gets printed, so the admin's own line endings elsewhere are untouched. Verified across CRLF, trailing-whitespace and clean markers, plus the no-file, prior-with-markers and prior-without-markers paths — six cases, all splicing correctly with admin content above and below the block intact.

Same commit adds a `!skip` guard on the begin rule. From the partial-marker state the function is documented to tolerate ("a previous run died mid-write"), a second `BEGIN` seen while already skipping re-fired the rule and emitted a duplicate of the whole generated block, which then sat in the vhost permanently. It now collapses back to a single clean block on the next run, stable over four consecutive runs.

### Still unverified on nginx

Two things I couldn't reach from a harness, both needing a real install run:

- the ordering at the `endManagedVhost` call — that the splice lands before `nginx -t` reads the file, so the test sees the spliced result and not the scratch copy
- `diffconfig()` actually firing on the nginx path now that the `mv -fv` backup is taken

### Unrelated, but it'll bite you on merge

The branch is three commits behind `working-1.6`, and two of those three are the pre-commit hook fixes. Consequences:

- `messages.pot` here still has 66 wrapped entries; on `working-1.6` it's fully unwrapped since the `--no-wrap` change. Those all churn when this merges.
- The older hook silently `return`s when gettext is missing, so translation regen appears to have been skipped for most of this branch's history — my first commit here swept up two uncaptured strings (`Use the default kernel`, `Use the default init`) that your PHP added a while back. Making that skip visible is the other of the two commits you're missing.

Merging `working-1.6` in picks up both and normalizes the `.pot` in one go, and it's only three commits, so probably worth doing before the divergence widens.

darksidemilk:
Closing as superseded by #1016, which folded this branch in and replaced the new-root three-zone design with additive intermediates hung off the existing `FOG Server CA` — the shape that reaches servers already in the field without re-pinning `ca.cert.der`. The update-workflow half landed via #1014.

Verified against the current `working-1.6` tip with git rather than the PR diff:

- **61 of the 63 commits on `pki-three-zone-phase1` are literal ancestors of `working-1.6`** (merge base `55df7e90e`).
- The **2 remaining commits** are not in by SHA but their content is byte-identical on `working-1.6`, re-applied on `pki-additive-intermediates`:
  - `6ffaba3b9` — nginx HTTPS redirect scoped to `location /` so the iPXE exclusion is reachable (`lib/common/functions.sh`, with the `location ^~ .../service/ipxe/` block intact).
  - `48f28b648` — CRLF/trailing-whitespace marker normalization plus the `!skip` guard in `spliceManagedBlock`.
- Every PHP method this PR added (`fogbase`, `fogpage`, `fogconfigurationpage`, `hostmanagement`, `groupmanagement`) exists on `working-1.6`; `bin/restorekernel.sh` and `lib/common/update.sh` are identical. Kernel/init dropdowns, kernel generations, the FOG-managed vhost block, iPXE background preservation, the `bin/setupacme.sh` removal, and the plan/spec docs all landed.
- Comparing every line each branch emits into the vhost (270 vs 278), exactly one differs, and `working-1.6` is the newer side: `ssl_certificate ${sslfullchain:-$sslpubcert}` instead of `ssl_certificate $sslpubcert`, plus a `dropShadowingVhosts()` this branch never had.

Not carried forward, deliberately:

- The three-zone-only machinery that #1016 replaced — `createRootCA()`, `createClientIntermediateCA()`, `_resolvePkiMode()`, and the `--split-pki` / `--legacy-pki` / `--root-ca-*` / `--client-ca-*` flags.
- The `custom.ipxe` pre-boot hook, removed in 06926e58ba3216c3adb019bec302f20ca4eeca19 as out-of-scope stacked work. `docs/SUPPORTED_CUSTOMIZATIONS.md` on `working-1.6` reflects this ("there is no supported hook point for pre-boot customization yet"). Not being revived at this time; it is a self-contained ~25-line change if it is ever wanted.

The nginx gap called out here is unchanged and now tracked on `working-1.6`: the redirect fix was measured against real nginx, but the managed-block splice and the rest of the vhost work remain Apache-verified only.

---

PR #1025, opened 2026-08-13
Close the two seams a third-party plugin cannot reach around

mastacontrola:
Prerequisites for opening plugins up to authors outside this repo. Both changes
close a place where core had to be hand-edited on a plugin's behalf — which
works exactly once, for a plugin that ships in this tree.

Split out of the plugin-architecture work so it can land on its own: the page
fail-open below is live on installs today and shouldn't wait on an ADR.

## Pages now fail closed

A node absent from the permission registry resolved to `null`, meaning "no
check". So a plugin page whose author never fired `PERMISSION_REGISTRY_DATA`
was reachable by **every authenticated user at every verb** — view, edit and
delete alike — while the same plugin's REST classes covering the same data
were already denied.

`resolveApiPermission()` closed the API side previously; this is the page side
of the same stance, using the same `unmapped.<node>` sentinel. The registry has
no `unmapped` node, so `assertCanGrant()` refuses to issue it and only a holder
of `*` satisfies it — an administrator keeps working, a restricted role loses
the free pass, and a log line names the node so the plugin author knows what to
register.

`buildMainMenuItems()` resolves a permission per menu entry per page load, so
the diagnostic is bounded to one line per node per request.

## Secret fields are declarable

`Route`'s two secret-field tiers were hardcoded arrays, so the only way to keep
a plugin's credential column out of REST output was to write it into core —
which is why `'ldap' => ['bindPwd']` sat in `$sensitiveAlwaysFields`. No
third-party plugin could ever do that.

`API_SENSITIVE_FIELDS` lets a plugin declare its own. LDAP now does, and core
names the plugin nowhere.

Read both tiers through `Route::sensitiveFieldMap()`, never the properties
directly, or plugin-declared fields are skipped. `bootmenu.class.php` matters
most here — that endpoint is the unauthenticated one.

Also removes the now-redundant `'ldap' => 'ldap'` entry in `API_CLASS_ENTITIES`
and the stale comment justifying it: `_pluginEntity()` derives the same mapping
from what the plugin registers, and the fail-open branch the comment describes
was already fixed.

## Non-regression

Nothing on a stock install changes. Verified by enumerating every reachable
page node from core, plugin and report classes and diffing against the registry
(resolving class constants), `EXEMPT_NODES` and `NODE_ALIASES`:

**36 reachable page nodes, all covered, zero become admin-only.** All 14 shipped
plugins register their node. No step-311-style grant is needed.

The hook plumbing was verified too: `processEvent()` passes `fastmerge()`'d
arguments by value, so a listener's mutations only land because the array
*elements* are references. Confirmed that the reference survives `fastmerge`
and that auto-vivifying `$arguments['always']['ldap'][]` writes through.

`working-1.6` only — `dev-branch` has neither `Authorization` nor the tiers.

## Docs

`docs/plugin-development.md` gains the fail-closed rule with a `permData`
snippet, the two-tier secrets table, and both hooks in the events table. Its
constructor example is updated to `registerInstalled()` — it was still teaching
the hand-rolled `in_array($this->node, self::$pluginsinstalled)` guard that
helper replaced.


mastacontrola:
Rebrought onto `working-1.6` and unblocked.

This had gone `CONFLICTING` because `packages/web/lib/plugins/ldap/hooks/addldapapi.hook.php` no longer exists here — the plugins moved to [`FOGProject/fog-plugins`](https://github.com/FOGProject/fog-plugins) in #1029 while this PR sat open. My sequencing error: this was the stated prerequisite for that work and should have merged first.

That was not just a merge annoyance. This PR is one change across two files that must land together:

| Side | Change |
|---|---|
| core `route.class.php` | `$sensitiveAlwaysFields = []` — **removes** the hardcoded `ldap => bindPwd` strip |
| plugin `addldapapi.hook.php` | **adds** it back via the new `API_SENSITIVE_FIELDS` hook |

Landing the core half alone stops the LDAP bind password — a directory service account credential stored cleartext — being stripped from API payloads.

Resolved as:

1. The hook change ported to `fog-plugins` and released as [v1.6.2](https://github.com/FOGProject/fog-plugins/releases/tag/v1.6.2). It differs from v1.6.1 in that one file only.
2. The plugin file dropped from this diff, and **`FOG_PLUGINS_VERSION` bumped to v1.6.2 in this PR**. `downloadplugins()` runs immediately before `configureHttpd()`, so an installer run applies both halves in one pass — the pin is what restores atomicity.
3. `working-1.6` merged in; the only conflict was that modify/delete.

Verified after the merge: `$sensitiveAlwaysFields` is still emptied, core still fires `API_SENSITIVE_FIELDS` in three places, no doc references the deleted path, and both changed PHP files lint.

Residual risk worth knowing: someone who rsyncs web files without running the installer gets core without the plugin update. Narrow, but it is the general shape of every future core↔plugin change.

Issue #529, opened 2022-12-24
Make custom webroot work or remove from the installer


Sebastian-Roth:
See details in https://github.com/FOGProject/fogproject/issues/502#issuecomment-1362255028 and following comments

mastacontrola:
Made to work rather than removed — `6805a2c50` on `working-1.6`.

The installer always knew where it had put things; nothing downstream did. About
a dozen places carried a literal `/fog/`:

**Installer** — the nginx `location` / apache `LocationMatch` / `RewriteRule`
templates all matched `^/fog/`, so the vhost described a URL with nothing behind
it. `$webdirdest` is a *filesystem* path and is always `<docroot>/fog`, while
`$webroot` is the *URL* path; with the default they coincide, which is why
nothing ever linked them. The tree is now also published at the requested path —
and the removal of `${docroot}${webroot}` in that same function has always
expected that link to exist, it just was never created. Apache does not resolve
symlinks when matching `<Directory>`, so the published path gets its own block,
without which a bare `/mywebroot/` 403s. `FOG_WEB_ROOT` was seeded with a
literal `/fog/` regardless of where the files went, so the database disagreed
with the disk; `WEB_ROOT` now goes into `config.class.php` beside `WEB_HOST`.

**Web tier** — `BootMenu` read `FOG_WEB_ROOT` and overwrote it with `'/fog/'` on
the very next line. That is precisely why a custom webroot broke PXE booting
(#502). `basename()`, which collapsed a nested `/apps/fog/` to `fog`, is now a
trim. AltoRouter's base path was the literal `'/fog'`, so on a custom webroot no
API route matched at all and existing endpoints answered **501**. Five pages
built server and node URLs from a literal `/fog/`; they now use a shared helper
that takes a storage node's *own* webroot when addressing that node, since nodes
need not share this server's.

### Verified on two Debian 12 containers

One installed `-W /myfog/`, one with defaults. Identical response codes on every
probe:

| path | `/fog/` install | `/myfog/` install |
|---|---|---|
| `management/index.php` | 200 | 200 |
| `service/ipxe/boot.php` | 200 | 200 |
| `system/status`, `system/info` | 200 | 200 |
| `bandwidth/1` | 200 | 200 |
| `system/export` | 403 | 403 |
| `host/1` | 403 | 403 |
| `/` | 308 | 308 |

The unauthenticated API list had to become webroot-relative to get there. It is
the same five endpoints with the same exposure — only the prefix is dynamic —
[...]

PR #1042, opened 2026-08-15
Describe the API in OpenAPI, generated from FOG's own metadata

darksidemilk:
Adds a machine-readable description of the REST API, generated from FOG's own metadata, plus a page in the web UI that renders it with a working try-it console.

## Why

Nothing machine-readable has ever described this API. What exists is prose — `kb/integrations/api.md`, still tagged *in-progress*, and a 2016 blog post that downstream clients cite as the reference. Prose goes stale the moment a class or a route is added, so in practice anyone writing a client reads `route.class.php`.

## What

**`GET {webroot}/system/openapi`** — and **`GET {webroot}/swagger.json`**, the same handler under the name most tooling reaches for first — returns an OpenAPI 3.1 document built from the structures the router and model layer already read at runtime:

| Source | Gives |
|---|---|
| `Route::$validClasses` × `defineRoutes()` | paths and methods |
| `ReflectionClass::getDefaultProperties()` | `$databaseFields`, `$databaseFieldsRequired`, `$databaseTable` |
| `commons/schema-expected.php` | column type, nullability, length, enum values, defaults |
| `Authorization::resolveApiPermission()` | the permission each operation requires |

The reflection call is the same one `Route::create()` and `edit()` already make to decide what they will accept, and the permission is resolved through the same call the router makes when the request arrives — so documented and enforced cannot drift apart.

Sample of what it produces:

```jsonc
"name":        {"type":"string","maxLength":250,"x-fog-column":"pAlias"},
"description": {"type":["string","null"],"x-fog-column":"pDesc"},
"ADPass":      {"x-fog-sensitive":"list",
                "description":"Omitted from list responses; returned only on a single GET by id."}
// operationId: deleteHost   x-fog-permission: host.delete
```

**Generated per request, not shipped as a file** — and that is the point. `$validClasses` is mutated by the `API_VALID_CLASSES` hook and the sensitive-field lists by `API_SENSITIVE_FIELDS`, so a build-time file would describe the classes FOG ships with rather than the ones a given server actually exposes. Generating live means the document is true for the server serving it, plugins included.

Served unauthenticated alongside `system/status` and `system/info` so a client can discover the server before it has credentials. It exposes only *shape* — class names, field names, types, all already public in the source — and no data. Both route names are declared `null` in `API_ROUTE_PERMISSIONS` rather than left to the unknown-route fallback, so "public on purpose" is recorded rather than inherited from a compatibility stance. It sits behind `FOG_API_ENABLED` like the rest of the tree.

## The UI page

A new **API Documentation** node renders the spec with Swagger UI 5.32.13. In the web UI rather than on the docs site because that is what makes try-it work: same-origin against the very install the admin is looking at, showing that server's classes. A console on a static docs page has no server to call. It also means fog-docs needs no Quartz OpenAPI plugin, which was the alternative.

Page discovery is already automatic, so no registry entry was needed — only a menu entry and a permission. `apidocs` aliases to the `settings` registry node, since a full map of every class and field is the same kind of server information the settings pages already carry.

Two deliberate choices: the signed-in user's API token is **not** pre-filled (it would save one paste at the cost of writing a live credential into the DOM), and the spec URL travels in a data attribute so it is escaped once, by PHP, as part of the markup. When the API is disabled the page says where to switch it on rather than rendering an empty console.

### Dark mode

Swagger UI ships exactly one theme, light, with no dark mode and no switcher; the community theme packs stopped at 3.x in 2022. FOG has had a working dark mode since the AdminLTE 4 work — `data-bs-theme` on `<html>`, driven by the `fogTheme` cookie or the OS preference — so left alone this page would be the one white rectangle in an otherwise dark UI.

`css/swagger-ui-fog.css` keys off that same attribute, so flipping FOG's own toggle re-themes this page with it rather than needing a second control that could disagree. It covers panels, tables, code blocks and icons, and in particular the form controls, which are white-on-white in dark by default and would make try-it unusable rather than merely ugly.

## Honest limits

Documented in the class rather than papered over:

- **`Route::getter()`'s per-class response augmentations** are a hand-written switch with no metadata behind them (`image` gains `os`, `imagetypename`, `osname`, `storagegroupname`; 7 other classes similarly; the other ~40 fall through to a plain `get()`). Described in prose on the affected operations.
- **Per-class create/edit extras** — `host`'s `macs`/`primac`/`snapins`/`printers`/`modules`/`groups` — are hand-coded `isset()` branches. Likewise prose.
- **A column with no manifest entry** is typed `string` and says so. Omitting the field would understate the API; guessing would misstate it.
- **1.6 only.** 1.5 has no `commons/schema-expected.php`, so there is no type source to generate against.

A class in `$validClasses` whose model will not resolve is skipped rather than fatal — one broken plugin should not take the document down — and `Error` is caught alongside `Exception`, because a class naming a missing parent raises the former.

## Validation

Validates clean against `redocly lint`: **0 errors**. Also accepted by `swagger-cli` and by `openapi-generator` 7.24, which generated a working Python client from it (54 files, `count_host()`, `cancel_host()`, …) — which is the point of shipping a description at all.

Four `no-ambiguous-paths` warnings remain and are expected: `/{class}/search/{item}` and `/{class}/{id}/task` are ambiguous as *templates* but cannot collide, because AltoRouter constrains the id segment to `[i:id]`, so `search` never matches it and an integer never matches the `search` literal. OpenAPI path syntax has no way to express that; inventing paths the server does not serve would be worse than the warning.

Also exercised by a standalone harness (44 assertions) that stubs the collaborators but feeds the generator the **real** `schema-expected.php` and real `databaseFields` arrays, so the type parsing is tested against production data. Happy to contribute that as a proper test if you have a preferred harness — I did not see an existing PHP test suite to add it to.

## Things you may want to push back on

- **Swagger UI 5.32.13 is vendored** — `swagger-ui-bundle.js` 1.5MB plus `swagger-ui.css` 182KB, Apache-2.0, license file alongside. Vendored rather than CDN because FOG servers are routinely on isolated networks where a CDN reference is a blank page. It is a new dependency and that is a maintainer call; for scale, the already-vendored `datatables.min.js` is 2.65MB.

  RapiDoc was the first choice and was dropped on evidence: last release **2024-10-11**, twenty-two months ago, and it failed to render this document at all (`unable to load spec`). An unmaintained renderer predating most of the OpenAPI 3.1 tooling work is a poor thing to vendor or to put in front of administrators.

- **`Route::webrootbase()`** is a new public accessor, so `servers[].url` is built from the value the router anchors its own paths to rather than reconstructing it from the setting and risking the two disagreeing (GH-529).

- **`php-cs-fixer` was not installed locally**, so the pre-commit hook skipped PSR-2 formatting on the new PHP. Expect a formatting diff from CI.

- **Not yet run against a live server.** No FOG install was available — the harness proves the generation logic and the type parser, but not that `getClass()` resolves all ~50 real model classes on a running system. That is the first thing worth checking, and the most likely place a surprise hides.

Unrelated but noticed while reading: `image.class.php` declares `'size' => 'imageSize'` twice (lines 44 and 47). Harmless in PHP, but a sign nothing machine-reads that array today.


darksidemilk:
## Verified against a real server

FOGProject/fog-workflows#13 stands up a throwaway FOG install and calls the API. Pointed at this branch with the OpenAPI assertions required:

```
PASS  system/info 200
PASS  host list 200 with tokens
PASS  host list refused without tokens
PASS  system/openapi 200 / valid json / declares openapi 3.1.0
      393 paths, 53 schemas
PASS  swagger.json 200 / valid json / declares openapi 3.1.0
      393 paths, 53 schemas
PASS  both paths serve the same document

PASSED: 14   FAILED: 0
```

**53 schemas is the number that mattered.** It means `getClass()` resolved every model class on a running system — the one thing this could not be checked for without a server. A document that resolved almost nothing would still have returned 200 and still parsed as JSON, so the count is the only assertion that actually distinguishes working from quietly broken.

From the generated document:

| | |
|---|---|
| `info.version` | `1.6.0-beta.3365` (read from `FOG_VERSION`) |
| `servers[0].url` | `http://10.88.0.2/fog` (built from `FOG_WEB_ROOT`) |
| `Host` schema | 30 properties, `required: ["name"]` |
| tags | 52 |

The live run also caught something no fixture could: sensitive-field tiering flagged **`ADPassLegacy`** alongside `ADPass`, `productKey`, `pub_key` and `sec_tok`. My local harness only knew about the first four — the generator picked the extra one up because it reads `Route::sensitiveFieldMap()` at request time rather than from a list baked into it. That is the request-time argument working exactly as intended.

This closes the "not yet run against a live server" caveat in the description. Still outstanding: `php-cs-fixer` was not available locally, so expect a PSR-2 formatting diff from CI.

Issue #1096, opened 2026-08-16
working-1.6: the three iPXE provisioning gates were not updated for the $netbootproto split


darksidemilk:
> **Rewritten.** The original text asked for `$netbootproto` to be split from `$httpproto`. That is already implemented on `working-1.6` — I missed it because I checked against a 1.5.x tree. The real defect is narrower: the split landed in one consumer and not in three others. Everything below is verified against `working-1.6`.

## What already works

The three-zone PKI work split netboot's protocol from the global one:

| Piece | Location |
|---|---|
| `--netboot-proto http\|https` flag | `bin/installfog.sh:503`, applied `:769` |
| `_resolveNetbootProto()` | `lib/common/functions.sh:4952` |
| `default.ipxe` uses `${netbootproto}` | `lib/common/functions.sh:1534` |
| Persisted as a managed key | `lib/common/functions.sh:4231` |
| Apache/nginx redirect exclusion when the two differ | `lib/common/functions.sh:5900`, `:6100` |

```sh
_resolveNetbootProto() {
    [[ -n $netbootproto ]] && return 0
    if [[ $httpproto == https && $externalca != yes && $caCreated != yes ]]; then
        netbootproto="http"
    else
        netbootproto="$httpproto"
    fi
}
```

So on a fresh HTTPS install using FOG's own CA, netboot already drops to plain HTTP by default.

## The defect

Three places still decide iPXE provisioning from `$httpproto`:

| Line | Function | Behaviour when `httpproto=https` |
|---|---|---|
| `:1630` | `downloadipxe()` | skips downloading the prebuilt release asset |
| `:1701` | `downloadipxesecureboot()` | **skips staging the Secure Boot chain** |
| `:1728` | `configureTFTPandPXE()` | compiles iPXE locally with `TRUST=` |

On the split's own **default** configuration — `httpproto=https`, `netbootproto=http` — all three take the HTTPS branch even though iPXE never opens a TLS connection:

1. Two clones and eight `make` invocations, no warm path, producing a binary whose baked-in CA is never used.
2. **The Secure Boot chain is silently absent.** The message is `Skipped (not usable with HTTPS)`, and the comment above it still reads:

   > These are upstream's generic signed binaries, so they cannot carry this server's CA — which is the whole reason an HTTPS install rebuilds iPXE locally. A signed binary cannot be rebuilt without invalidating the signature, so Secure Boot and HTTPS are mutually exclusive here.

[...]

PR #1097, opened 2026-08-16
Docs: transport modernisation spike for the boot path and image path

darksidemilk:
Adds one document. No code, no behaviour change.

`docs/superpowers/specs/2026-08-16-boot-and-image-transport-spike.md`

## Why

#1043 asserted in passing that HTTP is the preferred transport over TFTP for PXE artifacts. This writes up what that would actually mean — and, since transports were on the table anyway, whether the image data path is leaving anything on the table.

The two halves are independent workstreams that would ship separately. They are in one document because they were investigated together, not because they belong together.

## Part A — boot transport

The finding that reframes the question: **hops 4 and 5 of the boot chain already run over HTTP/HTTPS** (`functions.sh:824`, `bootmenu.class.php:392`). Hops 1–3 carry ~1 MB of NBP plus a 700-byte script, so there is no throughput argument for moving them. The real arguments are reachability, dependency reduction, and firmware that has HTTP Boot but no working PXE ROM.

Covered:

- UEFI HTTP Boot is a **separate firmware boot entry**, not a mode of the PXE entry — and Dell/Lenovo support a **static boot URI** that overrides DHCP, so it can be piloted with zero DHCP change.
- A side-by-side DHCP option table (opt 60/66/67/93) against exactly what `_keaBaseClasses()` emits today, so "what changes" is legible at a glance.
- **proxyDHCP was the gating unknown and it works** — `dhcp-pxe-vendor` is the missing piece; without it dnsmasq stays silent in proxy mode because it does not recognise `HTTPClient` as a PXE vendor class.
- Why `/tftpboot` should stay primary rather than inverting the tree under the web root: symlinks are already rejected for TFTP-served content (`functions.sh:1031-1044`), the two trees carry different SELinux labels with no policy bridging them, and `_publishSecureBootKit()` already answered this by copying.
- Attack surface: the trust model does not change (Secure Boot is the control, and it is transport-independent); what grows is **pre-boot firmware** surface, with PixieFail's TCP-ISN and weak-PRNG CVEs landing specifically on the TCP+DNS that only HTTP Boot uses.
- Why UEFI **HTTPS** Boot stays out of scope — firmware ships no public root store, so Let's Encrypt buys nothing there.

## Part B — image transport

**partclone constrains nothing.** It reads a stream and FOG already isolates it behind a FIFO, so the entire unicast deploy transport is one `cat` in `writeImage()`. The multicast path already proves the seam works. The binding constraint is the metadata sidecars, not the bulk data — and the `*FileName()` helpers already centralise those.

Frames transport as optional config: one out-of-the-box default plus two tiers of opt-in (Tier 1 against FOG's own NFS server, Tier 2 against infrastructure the admin brings). Identifies one shared prerequisite gating every NFS improvement at once — **BusyBox cannot mount NFSv4 at all and cannot express `nconnect`**, so `nconnect`, NFSv4.1/4.2 and pNFS are unlocked together or not at all.

Flags `rsize=32768` in `fog.mount` as the first thing to measure, with ADR-0013 as both the precedent and the warning.

## Related

- **#1096** — filed from this work: decoupling the netboot protocol from the global `httpproto`, so HTTPS can become the default. Includes the finding that `downloadipxesecureboot()` skips the signed chain on HTTPS installs, making Secure Boot and HTTPS mutually exclusive today.
- A second finding — the vhost-scoped PHP `SetHandler` applying through symlinks — is noted in the doc but belongs to #1043 and is being handled there.

## Note on branch target

Based on `dev-branch`; the diff is one new file. The `docs/superpowers/specs/` convention actually lives on `working-1.6`, which is also where #1043 sits, but the two branches share no usable history (17k/518 commits apart) so this could not target there directly. Happy to port it as a separate PR if `working-1.6` is the preferred home.

## Verification

Documentation only — nothing to run. Every code reference in the document was checked against the tree; claims that rest on inference rather than verification are marked as such inline, and the open questions and the hardware testing each part would need are listed at the end.

https://claude.ai/code/session_01CPVvfYojGn871mgmYegqmX

---
_Generated by [Claude Code](https://claude.ai/code/session_01CPVvfYojGn871mgmYegqmX)_

### DIFF
--- a/docs/superpowers/specs/2026-08-16-boot-and-image-transport-spike.md
+++ b/docs/superpowers/specs/2026-08-16-boot-and-image-transport-spike.md
@@ -1,0 +1,576 @@
+# Transport modernisation spike: boot path and image path
+
+**Status:** investigation spike — no implementation proposed yet
+**Date:** 2026-08-16
+**Prompted by:** [#1043](https://github.com/FOGProject/fogproject/pull/1043)
+
+---
+
+## Why this exists
+
+PR #1043 (Secure Boot signing for local ESP boot) asserted in passing that HTTP is the preferred
+transport over TFTP for PXE artifacts. That raised two questions worth answering properly:
+
+1. How much of FOG's boot chain could move to HTTP or HTTPS, and what would that cost admins?
+2. Once transports are on the table at all — is the image data path, NFSv3 since forever, leaving
+   performance or security on the table?
+
+**These are independent workstreams.** They share only the word "transport". They would ship as
+separate PRs, on separate schedules, and neither depends on the other. They are in one document
+because they were investigated together, not because they belong together.
+
+Nothing here is a proposal to change code. Every finding below is either verified against the
+tree or explicitly flagged as an open question.
+
+### One correction to #1043
+
+The PR says "iPXE's docs favour it over TFTP." ipxe.org does not state that as a documented
+recommendation — that was checked. The substance holds (HTTP is faster for bulk transfer and is
+a large part of why iPXE exists), but the appeal to authority was thinner than written and
+should not be repeated.
+
+---
+
+# Part A — Boot transport
+
+## FOG is already almost entirely HTTP
+
+This is the first thing to understand, because it reframes the whole question.
+
+| Hop | Transport |
+|---|---|
+| 1. DHCP → `next-server` + `filename` | TFTP |
+| 2. Firmware fetches `snponly.efi` / `undionly.kkpxe` | TFTP |
+| 3. iPXE embedded script → `default.ipxe` | TFTP |
+| 4. `default.ipxe` → `boot.php` | **HTTP/HTTPS already** |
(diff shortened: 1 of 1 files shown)


darksidemilk:
Closing this — it was opened against the wrong base and the content is moving elsewhere.

**Wrong base.** The branch was cut from `stable`, not `dev-branch` or `working-1.6`. `stable` is the release line and should never be a development base.

**Wrong home.** On reflection this is exploratory research rather than a spec, so it does not belong in `docs/superpowers/specs/` on a code repo. It now lives in `fog-docs` under `research/`, a directory outside `docs/` that Quartz never builds — versioned, but not published to docs.fogproject.org.

Re-verifying the document against `working-1.6` while moving it turned up something that changes the picture: **`working-1.6` has already split `$netbootproto` from `$httpproto`** as part of the three-zone PKI work. The three iPXE provisioning gates were not updated to match, which is a real and narrower bug — #1096 has been rewritten around it.

No changes are lost; nothing from this branch is being merged into `fogproject`.

---

PR #1109, opened 2026-08-16
Sweep gettext msgids that were built at runtime, and gate it

mastacontrola:
`xgettext` extracts the **literal source text** of a `_()` argument, but PHP evaluates that argument before `_()` is ever called. Where the two differ, the msgid sitting in the catalog and the string looked up at runtime are different strings, and the lookup can never hit.

```php
echo _("Add selected $getType");

// msgid in messages.pot:  "Add selected $getType"
// looked up at runtime:   "Add selected Image"
```

gettext misses, returns its argument unchanged, and the string renders **in English on every install whatever language is selected**. Nothing errors and nothing logs. The only symptom is a string that is never translated — invisible to anyone testing in English, which is why so many had accumulated.

This is the sweep of the bug class #1108 fixed one instance of.

## The fix

Same shape throughout: the literal goes **inside** `_()`, the variable is substituted **outside** it. That is already the house style — `openapi.class.php` uses it a dozen times and `$foglang['ListAll']` is literally `_('List All %s')`. It hands `xgettext` a complete format string and hands the translator a placeholder they can move, which matters: several shipped languages cannot keep English's word order.

Two sites needed something else:

- The submenu builder wrapped `_()` around a `sprintf` whose argument had **already been translated** a few lines above. The outer call was a guaranteed miss returning its own argument, so it is simply removed — a no-op at runtime.
- `fogservice.class.php`'s `_("{$itemType}s")` becomes `_($itemType) . 's'`, matching the `_($itemType)` already sitting three lines away in the same expression. Same rendered output, and it can now resolve.

## The gate

`tests/gettext-literal-msgid.test.php` draws its line at **does this msgid mix literal text with a variable**. It deliberately still *allows* a bare runtime lookup like `_($node)`: that resolves correctly whenever the value is itself a msgid extracted from some other literal site, which is how FOG translates node and class names. Fragile, since `xgettext` cannot verify it — but not broken by construction the way the mixed form is. The distinction needs no allowlist: an argument either carries literal text or it does not.

Tokenizer-based, for the same reason `php-deprecations.test.php` is: a regex cannot tell an interpolating `"…"` from a non-interpolating one, and cannot tell `_(` the gettext call from `->_(` a method.

## Verification

The proof is the catalog the pre-commit hook regenerated, not an assertion — every new msgid is now extractable where the composed forms never were:

```
All %s
Confirm you would like to delete this %s
Remove %s Associations
Please confirm you would like to dissociate the selected %s
Export %s
GPU-%d Vendor / GPU-%d Product
This machine already registered as %s
Unsafe hostname entered, please try again: %s
There are no %s on this server
Interface Ready with IP Address: %s
```

`php -l` clean on every changed file; the new gate passes on both branches.

## Not fixed here

`fogservice.class.php` builds its log lines by concatenating separately-translated fragments (`_('Not syncing')` + type + `_('between')` + …). That is a real i18n problem, but rewriting those into whole sentences changes daemon log output people grep, so it wants its own change.

## Downstream

None. No route, no schema, no API surface — `openapi.class.php` untouched by design.

https://claude.ai/code/session_017aBSWrDArXHTpKWkkN27LR

### DIFF
--- a/packages/web/lib/fog/fogpage.class.php
+++ b/packages/web/lib/fog/fogpage.class.php
@@ -815,14 +815,16 @@ abstract class FOGPage extends FOGBase
         $refNode = _($refNode);
         $menu = [];
         $menu = [
+            /**
+             * No _() around the sprintf: $refNode was already translated on
+             * its own above, and a msgid built at runtime can never match the
+             * literal xgettext extracted -- so the outer call was a guaranteed
+             * miss that returned its own argument. Dropping it changes nothing
+             * at runtime and stops the line claiming to be translatable.
+             */
             'list' => sprintf(
                 self::$foglang['ListAll'],
-                _(
-                    sprintf(
-                        '%ss',
-                        $refNode
-                    )
-                )
+                sprintf('%ss', $refNode)
             ),
             'add' => sprintf(
                 self::$foglang['CreateNew'],
@@ -998,7 +1000,7 @@ abstract class FOGPage extends FOGBase
             if ($node == 'ipxe') {
                 $this->title = _('All Boot Menu Items');
             } else {
-                $this->title = _('All ' . $this->childClass . 's');
+                $this->title = sprintf(_('All %s'), $this->childClass . 's');
             }
             $this->indexDivDisplay();
         } else {
@@ -2716,7 +2718,7 @@ abstract class FOGPage extends FOGBase
             _('Delete')
             . ': '
             . \Initiator::e($this->obj->get('name')),
-            _("Confirm you would like to delete this $node")
+            sprintf(_('Confirm you would like to delete this %s'), $node)
             . $extra,
             self::makeButton(
                 'closeDeleteModal',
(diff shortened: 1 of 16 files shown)


mastacontrola:
### Demonstrated against a real compiled catalog

Rather than assert the behavior, here it is. A French catalog containing **both** spellings — the composed msgid *and* the literal one:

```po
msgid "Add selected %s"
msgstr "Ajouter le %s selectionne"

msgid "Add selected $getType"
msgstr "SHOULD-NEVER-BE-USED"
```

Compiled with `msgfmt`, bound, `LANGUAGE=fr_FR`, `$getType = 'Image'`:

```
OLD  _("Add selected $getType")   -> Add selected Image
NEW  sprintf(_('Add selected %s'))  -> Ajouter le Image selectionne
```

The old form renders **English even though a French translation is present**, because the key it looks up is `"Add selected Image"` — which is neither entry. Note the literal-source entry is never used either, which is why simply adding the composed string to the catalog by hand would not have fixed it.

### Rendered English is unchanged

All 22 converted expressions were evaluated in both forms against representative values; every one is byte-identical, so this is a translation fix and not a copy change:

```
ok  All <class>s          => All Hosts
ok  delete confirm        => Confirm you would like to delete this host
ok  assoc modal body      => Please confirm you would like to dissociate the selected Snapins
ok  export title          => Export Hosts
ok  GPU vendor            => GPU-2 Vendor
ok  itemType plural       => groups
ok  advanced actions      => Group Advanced Actions
...  22/22 identical
```

Issue #1121, opened 2026-08-17
Phase 4: build out the certificate management page


darksidemilk:
Part of #1116. Last and separate — **explicitly out of scope for Phases 1–3**, its own branch and phase.

## What exists today

`FOGConfigurationPage::certificates()` (`packages/web/lib/pages/fogconfigurationpage.page.php:340`) is **read-only**: it shows the certificate hierarchy, explains that FOG uses certificates for three unrelated jobs, displays the trust anchor's SHA-256 and SHA-1 fingerprints, and links to the enrolment kit.

Everything an admin might actually want to *change* is currently done by re-running the installer with flags, or by hand-editing `.fogsettings`.

## What to add

- **Upload an external root CA** — the step-ca / corporate-root case. Should land it where `--web-ca-root`/`--ca-root` would have, and get it into the server's OS trust store the same way `_installCATrustAnchor()` does.
- **Download the individual public certificates in the chain** — root, intermediate, leaf — rather than only the enrolment kit's `ca.cert.der`.
- **Swap in the admin's own PKI**, with the guardrails the installer already encodes (a rotated signing key silently invalidates every existing enrolment; the Web CA key is FOG's whatever the leaf's provenance).
- **Set the PKI-related `.fogsettings` keys from the UI** — `publicWebCert`, `acmeLeaf`, `httpsRedirect`, `rebuildIpxeWithMyCA` — so the choices Phase 2 introduces are reachable without re-running the installer.

Link out to the Let's Encrypt worked example from #1120.

## Care needed

These settings decide whether FOG regenerates the admin's certificate and whether it locks the private key away from a renewal process. Getting one wrong from a web form is more destructive than getting it wrong on a command line, so the page needs to state the consequence at the point of change — particularly that `acmeLeaf`/`publicWebCert` are what stop FOG re-issuing a leaf it did not create.

darksidemilk:
## Sub-CA delegation of the FOG root — design notes, handed over from #1118

Surfaced while auditing `httpproto` coverage for #1118 and deliberately left out of both Phase 1 and Phase 2 (#1119). Recording it here because it belongs with the certificate-management page, and because most of the machinery already exists as dead code — which is easy to rediscover and misread as "nearly done".

Longer form in `docs/HTTPPROTO_COVERAGE_AUDIT.md` §A5 on `working-1.6`.

### What it is

[...]

Issue #1120, opened 2026-08-17
Phase 3: document netboot transport, PKI choices and Secure Boot enrolment


darksidemilk:
Part of #1116. **Blocked by #1119** — written against shipped behaviour, not intent. Work lands in [FOGProject/fog-docs](https://github.com/FOGProject/fog-docs).

Testing overturned several things the current docs imply, most importantly that HTTPS and Secure Boot conflict. They do not.

## New page

`docs/kb/reference/netboot-transport-and-pki.md`, organised around the four install modes, since the mode is now the thing an admin actually picks.

Repo rules apply: front matter from `docs/assets/templates/metadata-template.md` with a **unique `context_id`**, lowercase-dashed filename, **wikilinks on one line** (a wrapped one renders literally and the build does not warn), literal `→` rather than an icon shortcode, and Obsidian callout syntax rather than MkDocs admonitions.

## Must state plainly

- **The FQDN and public-DNS prerequisite** for `public-cert`, and *why*: a certificate is issued to a name, public CAs will not issue for a private IP, and the booting client must resolve that name from DHCP-supplied DNS — a prerequisite the IP-based URL never had.
- **`ca.ipxe.org` reachability** is what the public-certificate path depends on. iPXE's `CROSSCERT` fetch happens **on the booting client**, so an air-gapped network needs `embed-ca` regardless of how public the certificate is.
- **No mode disables Secure Boot.** Every mode stages upstream's signed chain and signs everything that does not arrive signed.
- **`embed-ca` *adds* an enrolment step** rather than removing one: the MOK must be enrolled before a client can netboot, because the chain has to hand off to a MOK-signed FOG build.
- **fog-client is what puts FOG's root into a client's trust store**, and what that means for machines that do not have it yet — this is the reason the HTTPS redirect is off by default.
- **The ESP `autoexec.ipxe` fallback covers which files were copied, not which driver works.** `chain X || goto Y` branches only when an image fails to *load*; once one loads and runs, control never returns. Do not let the page imply automatic driver selection.

## Flags and cross-links

Document `--install-mode` and every settings key in `docs/installation/server/command-line-options.md`. Note that **`--netboot-proto` has never been documented at all**.

[...]

Issue #1118, opened 2026-08-17
Phase 1: audit httpproto coverage and assess a conditional HTTPS redirect


darksidemilk:
Part of #1116. **Blocks Phase 2** — `httpproto` has to mean what the model says before anything is built on it.

## 1. Audit every non-netboot URL FOG emits

`httpproto` is about to become "the protocol FOG uses for its own **non-netboot** URLs", and its default moves to `https`. That is only safe if it is actually honoured everywhere.

- **Installer-side**: the `curl` calls in `lib/common/functions.sh` (node registration at `:508`, schema update, DB backup, node cert fetch, `create_update_node.php`), and any URL seeded into `globalSettings`.
- **PHP-side**: note that `FOGBase::$httpproto` (`packages/web/lib/fog/fogbase.class.php:491`) derives from `$_SERVER['HTTPS']` — i.e. **how the request arrived**, not the configured value. Decide per call site which of the two is correct; they are not interchangeable, and self-deriving is right for some of them.

Report anything that hardcodes a scheme or ignores the setting.

## 2. Assess a conditional redirect

A forced HTTP→HTTPS redirect must not drag netboot fetches onto HTTPS where they cannot validate — iPXE without a public certificate or an embedded CA will simply fail.

**Precedent already exists**: the nginx vhost excludes `service/ipxe/` from the redirect when `netbootproto != httpproto`. Determine whether that generalises:

- Does the Apache `RewriteRule` path need the same exclusion?
- Is `service/ipxe/` the complete set of netboot-reachable paths, or are there others (`boot.php`, the kernel/init fetches, `service/localboot/`)?
- Can this be expressed as a rule, or is an exclusion list the honest answer?

**Report the finding either way** — "exclusion list only" is a valid outcome and worth writing down rather than rediscovering.

## 3. Long-lived certificates must not be regenerated on every run

The leaf's historic test was `[[ ! -x $sslpubcert ]]` — true of every certificate ever written — so it was re-signed on **every single run**. Harmless while one key did every job, fatal once the signer can be offline. That one is fixed (re-issued only when the SAN set changes), but the class has never been audited.

Enumerate everything the installer writes under the PKI tree and classify each as **create-once**, **re-issued only on a real input change**, or **regenerated per run** — then fix anything long-lived in the last category.

## 4. Verify an external root reaches the OS trust store

[...]

Issue #1117, opened 2026-08-17
Phase 0: ESP boot binaries should be fog*, not local*


darksidemilk:
Part of #1116. Independent of the other phases — closes with #1043.

`_publishLocalBootFiles()` publishes a copy-verbatim ESP kit at `<webroot>/service/localboot/esp/`, in which FOG's own iPXE builds are renamed so they do not collide with the names shim reserves. They currently carry a `local` prefix.

**Rename them to `fog`**: `fogipxe.efi`, `fogsnp.efi`, `fogintel.efi`, `fogrealtek.efi`.

The prefix should name the binaries' *origin*, not one use of them — these have uses beyond local PXE boot, and `local` reads as though local ESP boot were the only thing they are for.

**The published directory stays `service/localboot/`.**

### Touches

- the `localbootespfiles` src→dst mapping in `lib/common/functions.sh`
- the generated `autoexec.ipxe`'s `chain` lines (`_espAutoexecScript`)
- `docs/superpowers/specs/2026-08-15-secureboot-local-esp-boot-design.md` §4
- `docs/SUPPORTED_CUSTOMIZATIONS.md`

### Do not rename

`snponly-shimx64.efi` and `snponly.efi` (and their aa64 equivalents). shim resolves its second stage by rewriting its **own** `-shim<arch>.efi` suffix to `.efi`, so `snponly-shimx64.efi` will load `snponly.efi` and nothing else, and it has to be upstream's copy because that is what shim's embedded certificate vouches for. Those two names are fixed by shim, not chosen by FOG — which is the entire reason FOG's own builds need a prefix in the first place.

### Verification

Re-run the publish loop against a mock tree and confirm 12 kit files with the new names, the two shim-reserved names unchanged, and the generated `autoexec.ipxe` chaining `fogipxe.efi` → `fogsnp.efi` → `fogintel.efi` → `fogrealtek.efi`.

darksidemilk:
Done — pushed to `ipxe-secureboot-symlinks` as `f23b9a8`, so this closes when #1043 merges.

- `localbootespfiles` now maps to `fogipxe.efi` / `fogsnp.efi` / `fogintel.efi` / `fogrealtek.efi`
- the generated `autoexec.ipxe` chains the new names, and its failure label is `:nofogbinary`
- `_signLocalIpxe()`'s header diagram corrected — it still showed `localipxe.efi` as the stage the signature is for
[...]

PR #1123, opened 2026-08-17
Phase 2: decouple the install settings, add --install-mode

### LINKED ISSUE #1119: Phase 2: decouple the install settings and add --install-mode
Part of #1116. **Blocked by #1118.** Supersedes the work already pushed to `claude/decouple-secureboot-httpproto`, which carries `_needsLocalIpxeBuild()`, unconditional Secure Boot staging, the private-CA caution, and addressing the server by name — all keep.

## Settings

See #1116 for the table and resolution rules. Each is an independent `.fogsettings` key.

- Split what `httpproto` conflates. `-S`/`--force-https` becomes purely `httpsRedirect`, which is what its name has always said.
- Rename `buildipxe` → `rebuildIpxeWithMyCA`, and make the build embed a **configured external CA** as well as FOG's own.
- Both `_createWebLeaf()` and `_hardenPkiPermissions()` guards become `[[ $acmeLeaf == yes || $publicWebCert == yes ]]` — OR, not an implication, so no setting silently mutates another.
- Firewall guidance (`functions.sh:2723`) lists 443 unconditionally; it is always listening.

## Migration

The case most likely to bite real servers:

- An existing `httpproto=https` is the only evidence its admin ever asked for `-S`, so **seed `httpsRedirect=yes` from it once**. Everyone else gets no redirect.
- `httpproto` moves to `https` for everyone. Safe because 443 already listens on every install and no redirect follows.
- Neither triggers a rebuild.

## Delete the openssl probe

`publicWebCert` is a persisted flag, full stop. The measurement that was prototyped is being removed for two reasons:

1. It was delicate — FOG adds its own CA to the host trust store by default, so a plain `openssl verify` answers "trusted" for FOG's *own* leaf, which is exactly the case that needs the rebuild. The configured CA has to be excluded first, and that exclusion must use **`-trusted`, not `-CAfile`**, because `-CAfile` *adds to* the default locations rather than replacing them. Both directions were verified against a two-CA fixture.
2. Decisively: a value re-derived every run from a trust store that other software also writes to is not something to hang a 25-minute build on. It has to persist.

## `--install-mode`

A **preset** that writes the keys — a convenience over the model, never a replacement.

| Mode | httpproto | netbootproto | publicWebCert | rebuild |
| --- | --- | --- | --- | --- |
| `standard` *(default)* | https | http | no | no |
| `http-only` | http | http | no | no |
| `public-cert` | https | https | **yes** | no |
| `embed-ca` | https | https | no | **yes** |

`public-cert` sets FOG's advertised web host to the FQDN. `embed-ca` must show a caution at selection time: it adds a Secure Boot enrolment step beyond the normal one (the MOK has to be enrolled before a client can netboot) and adds 10–25 min to every install **and update**, with no warm path.

## Attended presentation

The installer has no wizard — it is a summary plus one `read blGo` at `installfog.sh:998`. Add a prompt ahead of that presenting all four modes **with their explanations**, following the `read -p` style at `functions.sh:739`. Suppressed under `-y`/`--autoaccept`. Echo the resolved settings in the summary so an unattended run records them too.

## Verification

- Each mode yields the right tuple; each discrete key overrides its own field and nothing else; everything survives a re-run through `.fogsettings`.
- `netbootproto` steers to https from either trigger and stays http otherwise; forcing https with neither warns.
- Migration: `httpproto=https` in → `httpsRedirect=yes` and no rebuild; `httpproto=http` in → `httpproto=https` with **no** redirect. Check the emitted vhost for the 308/RewriteRule in each case.
- `embed-ca` with an external CA configured embeds **that** CA, not FOG's.
- Every mode: `$tftpdirdst/secureboot/` is populated, and an un-enrolled machine still netboots to the menu and can still run *Enroll Secure Boot Key*.

darksidemilk:
Phase 2 of #1116, implementing #1119. Blocked by #1118, which merged as #1122.

`$httpproto` decided three unrelated things: what protocol FOG uses for its own URLs, whether HTTP redirects to HTTPS, and whether iPXE is recompiled with this server's CA embedded. Nothing in its name says the last two, and an admin passing `-S` was not choosing them.

Plan and reasoning: [#1119 comment](https://github.com/FOGProject/fogproject/issues/1119#issuecomment-5312614692). Decision record: `docs/adr/0015-install-settings-are-independent-keys.md`.

## What was broken, before any default moved

| Site | Effect |
| --- | --- |
| `downloadipxe()` | iPXE release asset never downloaded on an HTTPS install, so such a server had no pristine copy of the published binaries at all |
| `downloadipxesecureboot()` | **every `-S` install staged no Secure Boot binaries whatsoever** |
| `configureTFTPandPXE()` | 10–25 min cold rebuild on every install *and* every update, no warm path, usually reproducing identical bytes |
| copy loop | overwrote any of the ~55 shipped binary names unconditionally — exactly the set an admin replaces |
| `_resolveNetbootProto()` | emitted nothing at all; an admin downgraded to HTTP netboot had no way to learn it |
| nginx HSTS | emitted on `:443` in **both** arms, including plain-HTTP installs; Apache emitted none |

The reasoning that tied the first three together — iPXE validates TLS strictly and cannot be told to trust a private CA — is true. What did not follow was inferring it from *"the web UI uses HTTPS"*, which says nothing about the netboot transport or what the certificate chains to.

## Four of #1119's premises did not match `working-1.6`

Worth stating so nobody re-adds them:

1. **`buildipxe` is not a setting.** `git grep` on `working-1.6` *and* `dev-branch` finds only `buildipxesrc` (a path) and the script name. `rebuildIpxeWithMyCA` is a **new key**, not a rename; no `deprecatedKeys` entry needed.
2. **"Embed a configured external CA" was already done here.** `ipxetrust` is `$sslcachain` (else `$sslcapem`), passed as `buildipxe.sh`'s only per-site input — under `--external-ca` that already *is* the admin's CA. (`dev-branch` still hardcodes `.fogCA.pem`.) Confirmed and commented, not rewritten.
3. **The firewall line is `_firewallPortList()`**, not `functions.sh:2723`. The issue is right that 443 should be unconditional — both web servers emit their `:443` vhost in *both* arms, so it has always been listening.
4. **`-S`'s help text already described the new meaning**: `--no-force-https` is documented as *"serve both HTTP and HTTPS without redirecting"*. Only the implementation conflated.

## Migration — one guess, made once

An existing `httpproto=https` is the only evidence its admin ever ran `-S`, so `httpsRedirect` is seeded from it. Everyone else gets **no** redirect. `httpproto` then moves to `https` for all, which is safe because 443 already listens everywhere and no redirect follows.

The seeding is guarded on `httpsRedirect` being unset so it fires exactly once. Without that, an admin who turned the redirect off would have it turned back on by the next upgrade re-reading `httpproto` — and HSTS rides the same key, so that reaches browsers rather than just the server.

The default at the top of `installfog.sh` deliberately stays `http`: the migration runs after `.fogsettings` is sourced and must distinguish a *persisted* https from a defaulted one. Writing https at the default site destroys exactly that distinction and would seed a redirect onto every fresh install.

## The `caCreated` trap

`_resolveNetbootProto` dropped to HTTP only when `httpproto == https` **and** neither `externalca` nor `caCreated` was set. `caCreated` is a **persisted** key, so it is `yes` on every re-run of an existing server. That was harmless only while `httpproto` defaulted to `http` — the moment it defaults to `https`, the old test resolves `netbootproto=https` on **every upgraded install in existence**, which is precisely the configuration that cannot work behind a private CA. Replaced outright, and pinned by a test.

## The five iPXE-tree requirements

| Requirement | How |
| --- | --- |
| Custom iPXE not overwritten | Sidecar manifest of checksums FOG wrote; a file that no longer matches is skipped and **named** in the run's output |
| Rebuild only when newer | `.fog-ipxe-build` stamp holding the iPXE tag **and** the embedded CA's checksum |
| Custom-built iPXE in a subfolder | `stock/` keeps the pristine downloads; rebuilt binaries keep top-level names |
| Sign the admin's own binaries | Already true incidentally — now deliberate, commented and counted |
| Warn loudly on HTTP netboot | `_reportNetbootProto`, printed last so it is still on screen |

**A manifest rather than the `fogsum` xattr** the kernel path uses. That mechanism no-ops when `attr` is absent or the filesystem carries no extended attributes, degrading to "unknown" — correctly treated as "not modified". Kernels survive that because they are backed up regardless; here there is no backup, so a silent degradation means silently overwriting the admin's binary *while reporting it protected*, and a TFTP root is exactly the sort of path that lands on a mount without xattr support. `attr` is not installed in the dev container, which is how the gap surfaced.

**The CA half of the rebuild stamp is not optional.** `TRUST=`/`CERT=` bakes the certificate in, so `--recreate-ca`, an external-CA switch or a rotation must force a rebuild at an *unchanged* tag — otherwise iPXE goes on trusting a CA that no longer signs anything, and it fails at a PXE client with nothing server-side to explain it.

**Top-level names are kept deliberately.** FOG's own `dhcpd.conf` emits `snponly.efi` / `undionly.kkpxe` / `arm64-efi/snponly.efi`, and most sites point an external DHCP server's option 067 at those same names — so no DHCP configuration anywhere changes. `stock/` excludes `secureboot/`, and that exclusion is load-bearing: `_signLocalIpxe` prunes exactly `$tftproot/secureboot` and nothing deeper, so a copy of it under `stock/` would fall outside the prune and get a FOG signature added to Microsoft's and iPXE's signed shim and loader.

**Protection starts at the first run after upgrading** — before that there is no manifest, so a previously-replaced binary is overwritten once. Stated in `SUPPORTED_CUSTOMIZATIONS.md` rather than left to be discovered.

## HSTS

The sharper half of the redirect split. A browser that has seen `Strict-Transport-Security` refuses plain HTTP to that host for six months **from its own cache**; no server-side change reaches it. Emitting it whenever `httpproto` was https — which is now everywhere — would have made it the one setting an admin could not take back. It follows `httpsRedirect` now.

Servers that have **already** sent it stay pinned. That cannot be fixed server-side; it is documented rather than papered over.

## The prompt

`input.sh`'s "Would you like to enable secure HTTPS on your FOG server?" is **removed**, not repaired. It set `httpproto`, which no longer varies — and it was never the question it claimed: answering "no" also disabled Secure Boot staging, answering "yes" started a 25-minute rebuild, and it mentioned neither. Its `wiki.fogproject.org` link is also dead.

`promptInstallMode` asks once, before the existing confirmation, showing all four modes with their costs. `embed-ca` states its caution *at the point of choosing*. Skipped under `-y`, when stdin is not a tty, and when the admin already answered via any of the flags. Any unrecognised answer takes `standard` — there is no wrong input worth re-asking for when the safe default is what an unsure admin should get.

## Tests

`sh tests/run-all.sh` green. Two new harnesses, 52 assertions, both verified to fail against the pre-change code (24 and 7 assertions respectively) so they are guards rather than restatements:

- `tests/install-settings-resolution.test.sh` — migration (including that seeding does not re-fire on a third run), the four modes, netboot steering, `--netboot-proto` winning in **both** directions, the `caCreated` trap, the warnings actually firing, and that every new flag is both accepted *and* documented.
- `tests/ipxe-tree-preservation.test.sh` — no-clobber across three runs, `stock/`, and the rebuild stamp skipping an unchanged rebuild but not a newer tag and not a changed CA.
- `tests/vhost-netboot-exclusion.test.sh` gains five assertions for the redirect/HSTS/firewall gating.

`bash -n` clean on `installfog.sh`, `functions.sh` and `input.sh`.

## Not in this PR

Sub-CA delegation of the FOG root — [design written up on #1121](https://github.com/FOGProject/fogproject/issues/1121#issuecomment-5312617686). The per-node protocol problem (~12 PHP sites pair the request scheme with a node IP; there is no per-node column). The `/fog/` GH-529 hardcodes. Documentation ports to `fog-docs` (#1120, now unblocked).

## Caveat

No installer has been run end to end against this. `run_all_distros.yml` in `fog-workflows` clones `dev-branch`, so it does **not** exercise this branch as-is. The riskiest untested paths are the real `buildipxe.sh` invocation under `embed-ca` and the SELinux relabel over the new `stock/` directory.

## Downstream

`fog-workflows/.github/workflows/reusable_api_validation.yml:264` reads `${httpproto:-https}` — already compatible with the new default, no change needed.

## Note on closing #1119

`Closes #1119` is **inert** here: GitHub only honours the keyword for merges into the default branch, and this targets `working-1.6`. #1119 needs closing by hand after merge, as #1117 and #1118 were.


mastacontrola:
Heads up on a numbering collision rather than anything in the change itself: `docs/adr/0014-install-settings-are-independent-keys.md` here, and `docs/adr/0014-authentication-seams-in-core-identity-providers-as-plugins.md`, which landed on `working-1.6` in #1135, are both ADR 0014.

Different filenames, so git won't flag it and both would just land — the same way two schema steps numbered 322 did a while back. Since #1135 is already merged, could you renumber this one to **0015**? It's the file rename plus the reference in the PR body; nothing else in the diff points at the number.

Nothing else from me on the change — flagging it now rather than after the merge.

darksidemilk:
Good catch, and thanks for flagging it before the merge rather than after — renumbered to **0015** in 45c1cc717.

Worth noting it had already stopped being hypothetical: `working-1.6` got merged into this branch in the meantime, so both files were sitting in the tree together, exactly as you described. Git recorded two adds rather than a conflict and said nothing.

Your read on the scope was right. The file states its number nowhere in its own text, and nothing under `lib/`, `bin/` or `tests/` references it, so it was the rename plus the one line in the PR body. `docs/refactor-phase2-plan.md` does mention "ADR 0014" twice, but those point at yours — it's PR 2.6 of the auth refactor — so I left them alone.

`tests/run-all.sh` is green on the merged branch, and the new PR CI passes on 7.4 and 8.3.

---

darksidemilk:
## Added: local ESP boot published as per-arch archives with a manifest (`0dd54c61`)

Folded in from #1166 (now closed) so this is one PR. Reworks #1117. It belongs here rather than on its own branch off `working-1.6` because the `.fog-ipxe-manifest` fix below depends on `_copyIpxeTree()`, which this branch introduces in `f646f22b`.

### Why

`_publishLocalBootFiles()` published the same bytes twice — `localbootfiles` as a "menu" under the TFTP names, `localbootespfiles` as a "kit" under `esp/` with the `fog*` names. On a real server:

```
service/localboot:
  10secdelay  arm64-efi  esp  i386-efi  index.php
  intel.efi  ipxe.efi  realtek.efi  secureboot  snp.efi
```

arm64 in four places (`arm64-efi/`, `10secdelay/arm64-efi/`, `esp/arm64-efi/`, `secureboot/arm64-efi/`). The delay set complete in one list and a single file in the other. And with listing off, the only way to enumerate the directory was to already know the filenames.

The premise that two lists served two audiences did not survive contact. The *only* reason the kit needed different filenames is that shim reserves `snponly.efi` and `ipxe.efi` on an ESP — and those names are the strictly safer choice everywhere. One correctly-named folder serves both.

### What is published now

```
service/localboot/
  index.php  manifest.json
  fog-esp-x86_64.zip   fog-esp-x86_64-10sec.zip
  fog-esp-i386.zip     fog-esp-i386-10sec.zip
  fog-esp-arm64.zip    fog-esp-arm64-10sec.zip
```

Each archive holds one top-level directory named after itself: upstream's shim pair(s) and MokManager, FOG's five builds under `fog*` names, `autoexec.ipxe`, `README.txt`, `MANIFEST.json`, plus whatever enrolment material the server publishes.

**The delay variant is its own archive, which removes a step rather than adding one.** iPXE runs exactly the file called `autoexec.ipxe` and cannot be asked which set you want, so a combined folder made choosing mean renaming one file over another. Chosen at download time, the binaries in a `-10sec` archive carry the plain names and its single `autoexec.ipxe` is already correct.

**`manifest.json` is what makes curation honest.** The old directory curated by *omission* — which is how FOG's `snponly.efi` came to be unavailable at all. A manifest carries the same advice per file (`role`, `origin`, `fogSigned`, a prose `note`) without withholding anything to give it. Static, written at install time — deliberately not a PHP endpoint that lists the directory on request, which would be a directory listing with a traversal surface.

### Reversals from the original design

| Was | Now |
| --- | --- |
| FOG's `snponly.efi` excluded | published as `fogsnponly.efi`, tried **last**, caveat in its manifest note |
| `MOK.der` not in the kit | in every archive — MokManager enrols by browsing the ESP for a certificate, so shipping it without one was a dead end one screen later |
| no `.auth` blobs | `PK`/`KEK`/`db.auth` in every archive |
| "i386 has no Secure Boot chain to assemble" | **corrected** — see below |
| signing gated on `--no-secure-boot` | signing always runs; the flag declines enrolment |

**i386 Secure Boot works, and the docs said it did not.** Shipping the `.auth` blobs means a machine in Setup Mode writes this server's certificate straight into `db`, after which firmware verifies a signed `fogipxe.efi` **directly — no shim, no MokManager, no MOK**. Upstream signs no shim for ia32, which is why the original design concluded i386 had no Secure Boot path.

**`--no-secure-boot` declines enrolment, not signatures.** It now suppresses `MOK.der`, the `.auth` blobs, and therefore the *Enroll Secure Boot Key* PXE entry (`BootMenu` already gates that on `MOK.der` existing). Keys are minted and binaries signed on every server: a PE signature is inert with Secure Boot off and costs nothing, whereas an unsigned binary is useless the day anyone enrols, with nothing on the server able to fix it short of a re-install. Applied in `_ensureSecureBootPlatformKeys()` and `_publishSecureBootKit()` instead of by blanking `$secureBootKey`. Installer help text updated — it claimed the flag left the FOS kernels unsigned.

### Bug fix in this branch's own `f646f22b`

`_copyIpxeTree()` records a sha256 of every TFTP file so a later run can tell FOG's copy from one the admin replaced. `_signLocalIpxe()` re-signs those `.efi` **in place** afterwards and never re-stamped. On the *second* install of a Secure Boot server all 45 `.efi` compare unequal: FOG stops updating its own binaries and prints "Kept your own copies of these iPXE files" for every one, permanently.

`_restampIpxeManifest()` rewrites only the lines for files actually signed — an entry deliberately carrying the *original* sum for a file the admin really did replace is copied through untouched, so that file stays protected. This is why the localboot work rides on this branch: on `working-1.6` there is no manifest to re-stamp.

### Trade-off, recorded because it is real

**No individual binary has a URL any more**, so nothing published here can be a UEFI HTTP Boot target or an iPXE `chain` destination — `service/secureboot/mmx64.efi` shows that is a real use, since the PXE menu chains that single file over HTTP. Accepted because the archive serves the "prepare an ESP" case this feature exists for; publishing the loose set alongside is purely additive if the other case turns up.

### Testing

`tests/localboot-publish.test.sh`, **64 assertions, all passing**. Every mock file's content is its own path in the tree, so provenance is read out of the file rather than trusted — which pins the failure that is otherwise invisible on a server: a `-10sec` archive whose `fogipxe.efi` came from the plain tree surfaces only as "the delay does nothing" on a switch running STP.

Also covered: the manifest parses as JSON under an independent implementation (python, not the shell that wrote it) with every sha256 matching the published bytes; no BIOS artifact (`.kpxe`/`.usb`/`.iso`/`.lkrn`) and no EMBED-less `autoexec/` build leaks in; upstream keeps the two shim-reserved names; the HTTPS-only shape publishes six archives and correctly omits an `autoexec.ipxe` no loader could read; the README never describes a fallback chain its archive lacks; a server with no enrolment material still succeeds; an empty tree reports `Failed` rather than publishing silently; re-runs are stable; and the `.fog-ipxe-manifest` regression is pinned end to end.

Both archive writers exercised — `zip` and the `tar -czf` fallback — and the produced `.zip` verified with both `unzip` and PowerShell `Expand-Archive`, which is the path that motivated preferring zip over tar.gz.

Full suite: 8 failures, **all pre-existing** — confirmed identical against a clean worktree at `bdb137b23` (`api-server-owned-fields`, `autoload-core-wins`, `no-session-for-browserless`, `plugin-extension-points`, `route-result-wrappers`, `ipxe-tree-preservation`, `pki-idempotence`, `trust-anchor`). `pxe-secureboot-menu-gating.test.php`, the one most exposed to the opt-out change, passes.

**Not yet verified:** hardware (both enrolment routes, including the i386 Setup Mode path) and an SELinux-enforcing distro.

### Not done, deliberately

- **`bzImage`/`init.xz` are listed, not bundled.** 60–80MB per architecture, and they would not produce a working local boot on their own — FOS reads per-host, per-task kernel arguments that `boot.php` generates. The manifest's `kernels` section points at the existing `service/ipxe/` copies with sha256s, so it indexes everything fetchable for a local boot without moving a byte.
- **`.usb`/`.iso` not published.** A coherent but different feature (boot iPXE off a USB stick without touching the ESP), with its own layout question.

### Reviewer note

`zip` is added to the package list. It is a separate package from `unzip` on every supported distro and is named `zip` identically on apt/dnf/pacman/apk, so no per-distro alternatives list is needed. A server whose package install partly failed falls back to `.tar.gz`, and `manifest.json` names whichever file was produced.

No REST route or `Route::$validClasses` change, so no `OpenAPI::document()` update and no FogApi class-list sync.

https://claude.ai/code/session_01S4x6HMrU32XFUNEqMbstbV

PR #1191, opened 2026-08-18
Remove the EFI binaries v2.0.0-fog.8 stopped shipping

mastacontrola:
Follow-up to #1190, which retired `autoexec/` but not `10secdelay/`'s EFI files — and that is the half that actually breaks a client.

## The bug

Those `.efi` are EMBED-marked, and the root now carries an `autoexec.ipxe`. `efi_autoexec_network()` falls back to `/autoexec.ipxe` when the binary's own directory has none — and `10secdelay/` has none — so one of them downloads the script, never runs it (`first_image()` returns its embedded one), nothing unregisters it, and `initrd_load_all()` concatenates 2 KB of iPXE script ahead of `init.xz`:

```
VFS: Unable to mount root fs on "/dev/ram0" or unknown-block(1,0)
```

That is precisely the failure the EMBED-less change exists to make impossible, reintroduced by an upgrade, on a path that worked before the upgrade.

`_copyIpxeTree()` cannot clear them — it only ever writes, so anything a release stops shipping stays on disk indefinitely.

## Observed, not theorised

On a server that had just upgraded to `v2.0.0-fog.8`:

```
$ strings -a /tftpboot/10secdelay/snponly.efi | grep -c 'Sleeping 10 seconds'
1                                    # still the fog.7 EMBED build
$ strings -a /tftpboot/snponly.efi   | grep -c 'Sleeping 10 seconds'
0                                    # root is correctly EMBED-less
$ ls /tftpboot/10secdelay/autoexec.ipxe
No such file or directory            # so it falls back to the root
```

`_signLocalIpxe()` had re-signed the stale file and given it a fresh mtime, so it looked current.

## The fix

`_retireAutoexecDir()` becomes `_retireStaleEfiPaths()` and sweeps:

| Path | |
|---|---|
| `autoexec/` | and `stock/autoexec/` |
| `10secdelay/*.efi` | the loose files beside the BIOS builds |
| `10secdelay/{i386,arm64}-efi/` | and the same under `stock/` |

The BIOS files in `10secdelay/` stay — they are why the directory still exists, since legacy BIOS has no `efi_autoexec_load()` and so cannot take a `--boot-delay`.

An admin who deliberately placed their own `.efi` there loses it. Accepted, and said so in the comment: every EFI binary that can legitimately sit in that directory now is one that would panic the client booting it, and a file that cannot be booted safely is not worth keeping over one that can.

Also rejoins `_copyIpxeTree()`'s leading comment, which #1190 split mid-sentence by inserting these helpers through the middle of it.

### DIFF
--- a/lib/common/functions.sh
+++ b/lib/common/functions.sh
@@ -2104,36 +2104,6 @@ _preserveStockIpxe() {
     done
     errorStat $?
 }
-# Copy the staged tree into place WITHOUT destroying an admin's own binaries.
-#
-# The historic loop was `find -type f -exec cp -Rfv {} $tftpdirdst/{}`, which
-# overwrites unconditionally. A file whose name is not in the staging tree
-# survives -- nothing deletes it -- but any of the ~55 names FOG does ship was
-# destroyed on every single run. That is the whole set an admin is most likely
-# to have replaced: snponly.efi, ipxe.efi, undionly.kkpxe.
-#
-# The customization machinery that would have protected it covers only the WEB
-# tree; both backupPreservedCustomizations and restorePreservedCustomizations
-# hardcode $webdirdest/service/ipxe. The TFTP tree was assumed safe "by
-# construction", which holds for new names and is false for colliding ones.
-#
-# So: record the checksum of every file FOG writes here, and skip a destination
-# whose current checksum no longer matches what FOG last wrote.
-#
-# A sidecar manifest rather than the fogsum XATTR the kernel path uses, and the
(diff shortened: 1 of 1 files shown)


darksidemilk:
You may want to make sure that the installer still doesn't remove any custom built ipxe files by default. I was trying to avoid removing someone's ipxe file that already has their CA embedded, and trying to make it so rebuild of it only happens if there's been a change to ipxe version or to the ca they have configured that would need to be embedded. 
Granted with the removal of embed script and needing autoexec, that may also be a flag for rebuilding custom ipxe.

PR #1192, opened 2026-08-18
Stop the published iPXE tarball overwriting a local CA build

mastacontrola:
## What broke

`--rebuild-ipxe-with-my-ca` compiles this server's CA into the iPXE binaries with `CERT=`/`TRUST=`. It is the only thing that lets a PXE client fetch `boot.php` over TLS when the web certificate chains to a private CA, and it is a 10-25 minute build, so a stamp in the TFTP root records what was built and later runs skip it.

The stamp recorded an *intention*, not a fact — the iPXE release tag and the sha256 of the CA, nothing about the binaries. `downloadipxe()` unpacks the published tarball into the same staging tree the build writes to, on **every** run, deliberately (that is where the pristine copy for `stock/` comes from). So:

| run | what happens |
|---|---|
| 1 | unpack → build → stamp → TFTP root gets the CA-embedded binaries, HTTPS netboot works |
| 2 | unpack, and the published binaries land on top of the build → the stamp still matches on tag and CA, so the rebuild is skipped → the TFTP root gets the **published** binaries, which trust nothing |

Nothing in the install output said the build had been skipped; the only line about it was `Rebuild iPXE with your CA: Yes` at the top. The failure surfaced at a booting client as iPXE `Permission denied (https://ipxe.org/0216eb8f)`, which decodes to `EACCES_USELESS` from `crypto/x509.c` — *"found no usable certificates"* — because the binary had no trusted root to build a path to.

First install fine, every install after it silently broken.

Found on a live 1.6 lab server: `/tftpboot/snponly.efi` contained no FOG CA at all, while the CA-embedded binaries from the previous run's build were still sitting in `/opt/fog/ipxe/build/ipxe-efi/src/bin-x86_64-efi/`.

## The fix

The stamp gains a third field, `bin=`, the sha256 of `snponly.efi` **as it sits in the staging tree**, so it describes what is actually there rather than what was once intended. That makes "no build needed" mean "the built binaries are still staged", which is exactly the condition under which the unpack must be skipped — so `downloadipxe()` skips it there, and unpacks in every other state: a new release, a changed CA, a staging tree someone emptied, or a server that does not build at all.

An existing stamp has no `bin=` field and therefore cannot match, so the next run rebuilds once and repairs the server it was written on.

## Test

`tests/ipxe-build-stamp.test.sh`, auto-discovered by `tests/run-all.sh`. Pins both halves. Mutation-verified: reverting the `bin=` field fails `rebuild after the published tarball overwrites the build` and nothing else.

## Downstream

None. No route classes, no schema, no API surface.

### DIFF
--- a/lib/common/functions.sh
+++ b/lib/common/functions.sh
@@ -2026,11 +2026,25 @@ _resolveIpxeTrust() {
 # at an unchanged iPXE tag. A version-only check would skip that rebuild, and
 # the failure lands at a PXE client as a TLS error with nothing on the server to
 # connect it to the cause.
+#
+# The third field, bin=, is the sha256 of snponly.efi AS IT SITS IN THE STAGING
+# TREE, and it is what turns the stamp from an intention into a fact. Without
+# it the stamp said "a build was run for this tag and this CA" and nothing
+# more: downloadipxe() unpacks the published tarball over the staging tree on
+# every run, so the run AFTER a build found the published, CA-less binaries
+# waiting there, matched the stamp on tag and CA alone, skipped the rebuild,
+# and copied the published binaries to the TFTP root. Every HTTPS netboot then
+# died at boot.php with iPXE's "Permission denied" out of x509.c -- no trusted
+# root to build a path to -- while the install printed nothing at all about the
+# build having been skipped. Comparing the staged bytes is also what lets
+# downloadipxe() decide whether unpacking is safe.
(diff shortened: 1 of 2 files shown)


darksidemilk:
What part of ALWAYS SIGN ALL THE BOOTABLE FILES did @claude not understand when I added that fog setting option? Thank you for catching this.

PR #1199, opened 2026-08-19
Unify the .fogsettings transport and PKI key names, and audit the rest

darksidemilk:
Part of #1120 (Phase 3). Base is `working-1.6`, not `stable`.

Documenting the settings model is what made the naming indefensible, so this does the rename the issue's follow-up comment asked for, fixes the one place the product still asserts the claim Phase 3 exists to retire, and records the audit.

## The false claim, still rendered in the UI

`fogconfigurationpage.page.php` rendered *"Secure Boot PXE and HTTPS are mutually exclusive: on an HTTPS install the installer skips these binaries"* — true of a gate `downloadipxesecureboot()` no longer has, while `_reportNetbootProto()` prints the opposite at the end of the same install. The product contradicted itself and the half an admin was likelier to read was the wrong one. Three comments in `functions.sh` and one test rationale used the same dead gate as their justification; each guard is still correct, for a different reason.

## The rename

`.fogsettings` carried two conventions for one subject: the keys ADR 0015 added are camelCase, the ones beside them were lower-case run-together (`httpproto`, `sslpath`, `catrust`, `secureboot`, `externalca`, `extca*`, the `ssl*` paths). All camelCase now, and `managedKeys` is grouped by subject so the settings that decide one thing sit together.

**No aliases.** An alias means the file carries two spellings of one setting forever with nothing to say which is live. Each old name is in `deprecatedKeys`, `_migrateLegacySettingNames()` copies its value across once, and the next write drops the stale line.

**Where that migration runs is the correctness argument.** It is called between sourcing `.fogsettings` and `doOSSpecificIncludes`, because `config.sh` applies `[[ -z $secureBoot ]] && secureBoot=1` and the same for `caTrust`. Called any later, those defaults find the new name unset and overwrite an admin's `--no-secure-boot` / `--no-ca-trust` with the value they opted out of. For the same reason `installfog.sh` no longer pre-seeds `httpProto`/`externalCA` ahead of the source.

**Two lookalikes deliberately untouched:** PHP's `self::$httpproto` is a static property derived from the request, and `'sslpath'` is a storage-node `$databaseFields` name (`'sslpath' => 'ngmSSLPath'`), also an API POST field and a CSV column. `registerStorageNode()` sends `-d "sslpath=$(... $sslPath ...)"` — the field name on the left is the ORM's, the variable on the right is ours. Nothing under `packages/web` was renamed.

## Verification

The managed key set is provably the old one with the mapping applied: same 79 keys, no additions, losses or duplicates. The full shell suite matches its pre-change baseline test-for-test; the remaining failures are openssl/platform fixture issues that reproduce identically on pristine `working-1.6`.

Three assertions needed updating and each tested a literal spelling rather than behaviour — including one that required `webCertFile` to sit *alone on its line*, so re-wrapping the list failed a test about persistence for a reason unrelated to persistence. That one now checks membership with comments excluded, which is stronger than what it replaced.

**New:** `tests/settings-name-migration.test.sh`, 17 assertions, pinning `managedKeys`/`deprecatedKeys`/the migration against each other **in both directions** — a name migrated but not deprecated leaves both spellings, a name deprecated but not migrated discards the admin's value — and asserting the call site's position by line number. Mutation-verified: moving the call after `doOSSpecificIncludes` fails it.

## Also here

- `docs/FOGSETTINGS_AUDIT.md` — twelve findings, deliberately **not** fixed here. Two of them (the empty `-10sec` archives, and the test fixture that hid them) were fixed independently by 046fe81e9 while this was being written, and are struck through with a note on what shipped.
- `--boot-delay`'s help text was spliced into the middle of `--netboot-proto`'s, so `--help` read as though fetching `boot.php` were something `--boot-delay` did. Found by transcribing the output into the docs, which is the only reason anyone reads it end to end.
- Merged `working-1.6` for the ESP archive rewrite. One conflict, resolved in favour of the incoming code; its `_espLocalAutoexecScript()` and new test needed `bootdelay` → `bootDelay`, without which the delay would have silently stopped reaching the ESP script.

## Downstream

No route classes changed, so no FogApi sync is implied.


darksidemilk:
Rethinking this

PR #1230, opened 2026-08-19
DO NOT MERGE - smoke test for the new release-pin CI check

mastacontrola:
Throwaway draft. Opened only to fire a `pull_request` event so the `release pins resolve` job added in FOGProject/fog-workflows#26 can be watched running against a real branch. Closing and deleting the branch as soon as it reports.


mastacontrola:
Done its job — the `release pins resolve` check ran and passed on a real pull_request event, resolving both pins with their assets. Closing; nothing here was ever meant to merge.

PR #1244, opened 2026-08-20
fix(schema): drop the zero-date default so MySQL 8.0 can install (#1243)

### LINKED ISSUE #1243: Schema cannot be applied to a stock MySQL 8.0: zero-date default on fileDeleteQueue.fdqCompletedDate
The schema cannot be applied to a stock MySQL 8.0 server. One statement, in both the fresh-install path and the reconciler path:

```
SQLSTATE[42000]: 1067 Invalid default value for 'fdqCompletedDate'
  ALTER TABLE `fileDeleteQueue` MODIFY COLUMN `fdqCompletedDate` DATETIME DEFAULT '0000-00-00 00:00:00'
```

`packages/web/commons/schema.php:4174`, and twice in `packages/web/commons/schema-expected.php`.

## Why it fails

`'0000-00-00 00:00:00'` is not a legal `DATETIME` default under MySQL 8.0's stock `sql_mode`, which includes `NO_ZERO_DATE` and `STRICT_TRANS_TABLES`. MariaDB's default `sql_mode` does not, which is why every MariaDB leg is clean and only MySQL fails.

**Nothing covers it.** `1067` is on neither tolerance list — not `SchemaUpdaterPage::update()`'s `$skiperrs`, not `SchemaReconciler::$_skiperrs`. FOG does relax `sql_mode`, but only inside `DatabaseManager::_convertEngine()`, only to `NO_ENGINE_SUBSTITUTION`, and it restores the previous value immediately afterwards — the schema update runs under whatever the server's own mode is.

So on MySQL 8.0 the update throws, `$newSchema->save()` is never reached, the schema version is never recorded, and `DatabaseManager::establish()` 308-redirects every request to `?node=schema` — the same downstream damage as #1147, including the installer's pre-upgrade `maintenance/backup_db.php` dump being answered with a redirect and an empty body.

## Measured

`tests/schema-executes.test.php`, once #1242 fixed the harness that was hiding it:

| Server | fresh-install path | reconciler path |
|---|---|---|
| MariaDB 10.5.29 | clean — 282 statements, 68 tables, 1 collation | clean |
| MariaDB 11.8.8 | executes clean (collation split is #1152/#1240) | clean |
| **MySQL 8.0.46** | **1 of 282 fails** | **1 of 67 fails** |

## Why it has never been reported

Three layers of silence, all now removed or accounted for:

1. Nothing in CI executed the schema against a database at all until #1221.
2. That job piped its test through `tee` without `pipefail`, so it reported green whatever the test returned (#1241).
3. With the pipe fixed, the test itself was failing 12 statements for its own reasons, which would have buried this one in noise (#1242).

Also worth saying plainly: it is entirely possible nobody runs FOG on MySQL 8.0 today. The CI matrix includes it deliberately, though — "MySQL has never had the UCA-14.0.0 collations at all, so it catches the same class of failure from the other vendor" — so either the schema should run there or the project should say it does not support it.

## Fix

The column wants a legal default. Three options:

1. **`DEFAULT NULL`** and a nullable column. Truthful — "not completed yet" is genuinely absent, not a date — and portable everywhere. Needs whatever reads `fdqCompletedDate` to handle NULL, which is the thing to check before committing to it.
2. **Drop the default entirely.** `DATETIME` with no default is implicitly nullable with a NULL default; same as 1 with a smaller diff.
3. **Keep the zero date and add 1067 to both tolerance lists.** Rejected — that tolerates the statement failing rather than making it work, so the column keeps whatever definition it already had and the two servers stay divergent. Tolerance lists are for "already how we want it", not "cannot do it here".

Whichever is chosen needs a schema step as well as the step-0 edit, because the column already exists with the zero-date default on every MariaDB install. **Number it after #1240's step 342** — that PR is in flight and adding a step concurrently will collide.

The manifest (`schema-expected.php`) carries the same literal twice and must be regenerated or edited in the same commit, or the reconciler path stays broken while the updater path is fixed.

mastacontrola:
Closes #1243.

`fileDeleteQueue.fdqCompletedDate` was declared `DATETIME DEFAULT '0000-00-00 00:00:00'` in step 288. That is not a legal DATETIME default under MySQL 8.0's stock `sql_mode`:

```
ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,
ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION
```

It answers **1067**, which is on neither `SchemaUpdaterPage::update()`'s `$skiperrs` nor `SchemaReconciler::$_skiperrs` — so the whole update threw before `$newSchema->save()`, after which `DatabaseManager::establish()` 308-redirects every request to `?node=schema`. **FOG could not be installed on MySQL at all**, in either path. MariaDB's default `sql_mode` has never carried either flag, which is why this sat in the tree for years and only MySQL ever objected.

## The fix

`NULL` says the same thing. Nothing reads this column — `FileDeleteQueueManager` only ever *writes* it, in `cancel()` and `complete()`, both with a real timestamp — so a zero date here means "never completed", which is exactly what NULL means, without needing a permissive server to express it.

Option 3 from the issue (add 1067 to the tolerance lists) is rejected: that tolerates the statement *failing*, so the column keeps whatever definition it already had and the two vendors stay divergent. Those lists are for "already how we want it", not "cannot do it here".

**Step 288 is edited** so a fresh install never creates the bad default. Editing a historical step is safe — it has already run everywhere it was going to.

**Step 343 repairs installs that already ran it**, and clears the rows *before* altering the column. The other order asks the server to convert an illegal value in place, which is itself an error on a strict server. The `WHERE` uses `YEAR(…) = 0` rather than comparing against the literal, because a strict server rejects that literal *in a comparison* too — the obvious spelling would fail on exactly the servers this exists for.

The manifest carried the same literal twice and is fixed in the same commit. Leaving it would have fixed the updater path and left the reconciler path broken.

## Verification

`tests/schema-executes.test.php`, all three servers, **283 statements, 68 tables, one collation** each:

| Server | before | after |
|---|---|---|
| mysql:8.0.46 | 1 of 282 failed, **both paths** | ok |
| mariadb:10.5.29 | ok | ok |
| mariadb:11.8.8 | ok | ok |

Step 343 against an install built the old way (`scripts/background_scripts/verify_zerodate_repair_1243.sh`):

```
before   default '0000-00-00 00:00:00';  rows a,b zero-dated, c a real timestamp
after    default NULL;                   a,b NULL;  c untouched
re-run   0 rows changed, column unchanged
```

## Gate

`tests/schema-portable-collation.test.php` gains a zero-date check, next to the two collation ones it already carries — the same class of bug, where the server that would refuse the statement is not the server anyone is writing on. The executable version is `schema-executes.test.php` against a real mysql:8.0, which is where it was found; this gate is so it cannot come back without a database in the room.

Mutation-verified, all three killed: restoring the literal in `schema.php`, restoring it in the manifest, and leaving `FOG_SCHEMA` at 342.

## Not fixed here, and bigger than this

While building the step-343 fixture I could not insert a zero-date row on **stock MariaDB 10.5** either:

```
ERROR 1292 (22007): Incorrect datetime value: '' for column 'fdqCompletedDate'
```

`FOGController::save()` writes `''` for any unset optional field whose key does not end in `id` (fogcontroller.class.php, the `$val = '';` arm), and `''` into a DATETIME is 1292 on **any** strict server — MariaDB has defaulted to `STRICT_TRANS_TABLES` since 10.2.4, so this is not a MySQL-only concern. The schema declares **50 DATETIME columns**, 22 of them `NOT NULL`.

So this PR makes the schema *install* on MySQL 8.0. It does not establish that FOG *runs* there, and I would not claim it does. Filed separately with the evidence.

## Sequencing

Depends on #1240 (step 342) being merged, which it is — this is step 343 on top.

### DIFF
--- a/packages/web/commons/schema-expected.php
+++ b/packages/web/commons/schema-expected.php
@@ -86,13 +86,13 @@ return [
             ],
         ],
         'fileDeleteQueue' => [
-            'create' => 'CREATE TABLE IF NOT EXISTS `fileDeleteQueue` ( `fdqID` int(11) NOT NULL AUTO_INCREMENT, `fdqPathName` varchar(255) NOT NULL, `fdqStorageGroupID` int(11) NOT NULL, `fdqCreateDate` datetime DEFAULT current_timestamp(), `fdqCompletedDate` datetime DEFAULT \'0000-00-00 00:00:00\', `fdqCreateBy` varchar(40) DEFAULT NULL, `fdqState` int(11) NOT NULL, `fdqPathType` varchar(255) NOT NULL, PRIMARY KEY (`fdqID`) ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_general_ci ROW_FORMAT=DYNAMIC',
+            'create' => 'CREATE TABLE IF NOT EXISTS `fileDeleteQueue` ( `fdqID` int(11) NOT NULL AUTO_INCREMENT, `fdqPathName` varchar(255) NOT NULL, `fdqStorageGroupID` int(11) NOT NULL, `fdqCreateDate` datetime DEFAULT current_timestamp(), `fdqCompletedDate` datetime DEFAULT NULL, `fdqCreateBy` varchar(40) DEFAULT NULL, `fdqState` int(11) NOT NULL, `fdqPathType` varchar(255) NOT NULL, PRIMARY KEY (`fdqID`) ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_general_ci ROW_FORMAT=DYNAMIC',
             'columns' => [
                 'fdqID' => 'int(11) NOT NULL',
                 'fdqPathName' => 'varchar(255) NOT NULL',
                 'fdqStorageGroupID' => 'int(11) NOT NULL',
                 'fdqCreateDate' => 'datetime DEFAULT current_timestamp()',
-                'fdqCompletedDate' => 'datetime DEFAULT \'0000-00-00 00:00:00\'',
+                'fdqCompletedDate' => 'datetime DEFAULT NULL',
                 'fdqCreateBy' => 'varchar(40) DEFAULT NULL',
                 'fdqState' => 'int(11) NOT NULL',
                 'fdqPathType' => 'varchar(255) NOT NULL',
--- a/packages/web/commons/schema.php
+++ b/packages/web/commons/schema.php
@@ -4171,7 +4171,15 @@ $this->schema[] = [
     'ALTER TABLE `fileDeleteQueue` CHANGE COLUMN `fqdCreateDate` `fdqCreateDate` DATETIME',
     'ALTER TABLE `fileDeleteQueue` CHANGE COLUMN `fqdCompletedDate` `fdqCompletedDate` DATETIME',
     'ALTER TABLE `fileDeleteQueue` CHANGE COLUMN `fqdCreateBy` `fdqCreateBy` VARCHAR(40)',
-    "ALTER TABLE `fileDeleteQueue` MODIFY COLUMN `fdqCompletedDate` DATETIME DEFAULT '0000-00-00 00:00:00'",
+    // GH-1243: NULL, not a zero date. '0000-00-00 00:00:00' is not a legal
+    // DATETIME default under MySQL 8.0's stock sql_mode (NO_ZERO_DATE,
+    // STRICT_TRANS_TABLES) -- it is error 1067, which is on neither
+    // tolerance list, so the whole schema update threw and FOG could not be
+    // installed on MySQL at all. NULL says the same thing ("not completed
+    // yet") and every supported server accepts it. Editing this historical
+    // step is safe: it has already run everywhere it was going to, and step
+    // 343 is what repairs those installs.
+    "ALTER TABLE `fileDeleteQueue` MODIFY COLUMN `fdqCompletedDate` DATETIME NULL DEFAULT NULL",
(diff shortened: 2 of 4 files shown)


mastacontrola:
## Correction: the severity in this fix was overstated

The DDL change here is right and stays. The reasoning I gave for it was not, and
the wrong version is in the commit message, the PR body and my close comment, so
it is worth stating plainly.

**What I claimed:** that `DEFAULT '0000-00-00 00:00:00'` meant FOG could not be
installed on MySQL 8.0 at all — error 1067 under `NO_ZERO_DATE`, the schema
update throws, the version is never recorded, every request 308-redirects to
`?node=schema`.

**What is actually true:** `PDODB::_connect()` runs

```php
self::query("SET SESSION sql_mode=''");
```

unconditionally, on every connection (`packages/web/lib/db/pdodb.class.php:298`).
So the schema updater never sees `NO_ZERO_DATE`, or `STRICT_TRANS_TABLES`, or
anything else the server would otherwise enforce. Verified in a stock `mysql:8.0`
container: with `sql_mode` cleared, both the zero-date DDL and `INSERT ... VALUES
('')` into a `DATETIME` succeed and store `0000-00-00 00:00:00`.

I found this because `tests/schema-executes.test.php` connects with a plain PDO
and inherits the *server's* `sql_mode`, which is strict. That gap between the
test and the runtime is deliberate and worth keeping — but it means a failure
there is not automatically an outage, and I read it as one.

**So the real finding is not "MySQL is unsupported", it is that FOG disables the
server's own validation on every connection** — undocumented, unconditional, and
the reason 83 of 86 hosts on my own 1.6 server carry `0000-00-00 00:00:00` in
`hostLastDeploy` with nothing ever having complained. That is now #1245.

The DDL fix still earns its place: `NULL` says what the zero date was standing in
for, and DDL that stands up without FOG having switched the server's checks off is
better DDL either way. The claim about installability was wrong.

Issue #1152, opened 2026-08-17
71 tables inherit the server default collation, which changed meaning in MariaDB 11.4


darksidemilk:
Follow-up to #1147, which fixed the collation names FOG *states*. This is about the ones it does not state.

## The gap

`commons/schema.php` declares 71 tables as bare `DEFAULT CHARSET=utf8` with no `COLLATE`:

```
 71  DEFAULT CHARSET=utf8              <- inherits whatever the server default is
  2  DEFAULT CHARSET=utf8mb3 COLLATE=  <- pinned by #1147
```

What that inherits is not stable across versions:

| Server | `DEFAULT CHARSET=utf8` resolves to |
|---|---|
| MariaDB < 11.4, all MySQL | `utf8mb3_general_ci` |
| MariaDB >= 11.4 | `utf8mb3_uca1400_ai_ci` |

#1147 fixed the case where a *generated* manifest carried the 11.4+ name to servers that could not execute it. This is the other half: on 11.4+ these 71 declarations quietly mean something different from what they mean everywhere else, and nothing in the tree says so.

## Why it bites on an upgrade specifically

Upgrading MariaDB does not rewrite existing tables. So on a server upgraded across the 11.4 boundary:

- every table that already existed keeps `utf8mb3_general_ci`
- every table created afterward comes out `utf8mb3_uca1400_ai_ci`
- the schema is now split, and nothing reports it

The split is silent until a **varchar** join crosses it, at which point it is `Illegal mix of collations`. Most of the joins that would expose it are int-keyed and will never show it — `siteRoleGrants` -> `roleUserAssoc` joins `srgRoleID` to `ruaRoleID`, both `int(11)` — which is precisely why this would sit undetected until some unrelated query happened to join on a name column.

A fresh install on 11.4+ is self-consistent (everything is uca1400) and is not affected. It is the upgrade path that splits.

## Options, none of them obviously right

1. **Name the collation on all 71**, as #1147 did for the two it touched. Most explicit, largest diff, and makes the tree say what it means everywhere.
2. **Pin the server default** — ship `collation-server = utf8mb3_general_ci` in the FOG server's my.cnf from `installfog.sh`. Small change, but FOG then owns a global server setting it does not own today, which is a real overreach on a shared database host.
3. **Convert the schema wholesale to uca1400** on 11.4+ and accept a version-dependent schema. Consistent per-server, but gives up on one schema meaning one thing.
[...]

PR #1248, opened 2026-08-20
fix(orm): store NULL for an empty date, not '0000-00-00 00:00:00' (#1245)

### LINKED ISSUE #1245: PDODB clears sql_mode on every connection, so FOG stores values the server would reject
> **Rewritten 2026-08-20.** The original text of this issue said the `''` write
> was "an error on any strict sql_mode" and treated that as the bug. It is not
> reachable, because FOG clears `sql_mode` on every connection — which turns out
> to be the more interesting finding, and is what this issue is now about. The
> same correction has been posted on #1243 / #1244, whose severity I overstated
> for the same reason.

## FOG disables the server's own validation on every connection

`packages/web/lib/db/pdodb.class.php:298`, in `PDODB::_connect()`, immediately
after the `new \PDO(...)`:

```php
self::query("SET SESSION sql_mode=''");
```

Unconditional. Every connection, every entry point — web UI, API, the schema
updater, all eight CLI daemons. It is the only place in `packages/web/lib` or
`packages/service` that sets a session `sql_mode`, and nothing restores it.

An empty `sql_mode` switches off `STRICT_TRANS_TABLES`, `NO_ZERO_DATE`,
`NO_ZERO_IN_DATE`, `ERROR_FOR_DIVISION_BY_ZERO` and the rest. Statements the
server would reject are accepted, values it would refuse are coerced, and the
coercion is reported as a warning that nothing reads.

There is no comment on the line, nothing in the installer, and nothing in the
docs. I have not been able to find the commit's stated reason.

## What it is currently hiding

`FOGController::save()` writes the empty string for any unset optional field
whose key does not end in `id`. The field loop has three arms:

- key is exactly `id` → omitted, so the column auto-increments;
- key **ends in** `id`, optional, empty → written as `NULL`;
- **everything else**, empty → `$val = '';`

Into a `DATETIME`, `''` is error 1292 under a strict mode and a silent coercion
to `0000-00-00 00:00:00` without one. FOG always runs without one.

A survey of the tree (`commons/schema-expected.php` joined against each model's
`$databaseFields`) gives the reachable set:

- 20 tables carry date/time columns, **33** such columns
- 32 are mapped by a model field; `virus.vDateTime` has no model field at all
- **16 are optional and not auto-filled by `save()`'s `switch ($key)`**, so they
  can reach the `''` arm:

  `fileDeleteQueue.completedTime`, `hosts.deployed`, `hosts.sec_time`,
  `images.deployed`, `imagingLog.finish`, `inventory.deleteDate`,
  `multicastSessions.completetime` / `.senderstart` / `.starttime`,
  `snapinTasks.checkin` / `.complete`, `tasks.checkInTime` /
  `.scheduledStartTime` / `.stateChangedTime`, `userAuths.expire`,
  `userTracking.date`

**13 of those 16 columns are `NOT NULL`**, which matters for the fix — see below.

## It is not theoretical

My own 1.6 server, MariaDB **11.8.8**, server `sql_mode` =
`STRICT_TRANS_TABLES,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION`.
Read-only count of `0000-00-00 00:00:00` rows:

| column | zero | total |
|---|---|---|
| `hosts.hostSecTime` | 85 | 86 |
| `hosts.hostLastDeploy` | 83 | 86 |
| `tasks.taskScheduledStartTime` | 44 | 57 |
| `images.imageLastDeploy` | 26 | 29 |
| `inventory.iDeleteDate` | 4 | 4 |
| `multicastSessions.msCompleteDateTime` | 2 | 16 |
| `imagingLog.ilFinishTime` | 1 | 11 |

That is a server whose own configuration forbids every one of those values,
holding them anyway, because FOG asked it to stop checking. Nothing has ever
complained — which is the point.

## Why this is worth fixing rather than leaving alone

Not because anything is failing today. Because:

- **Every value FOG stores is unvalidated, not just dates.** Truncated strings,
  out-of-range numerics and divide-by-zero all pass the same way. Dates are
  simply where it is visible.
- **The dependency is invisible and load-bearing.** Anyone tightening the schema,
  or a distro shipping a stricter default, is safe only for as long as this line
  stays. Removing it — which looks like a one-line cleanup — breaks writes across
  eleven model classes at once.
- **It defeats the test suite by construction.** `tests/schema-executes.test.php`
  connects with a plain PDO and inherits the server's mode, which is why it
  caught #1243 at all. Anything checked through FOG's own connection is checked
  under no rules.
- **`0000-00-00` is not what the code means.** `hosts.deployed` empty means
  "never deployed". NULL says that; a zero date is a real timestamp that display
  and comparison code has to special-case, and mostly does not.

## Shape of the fix

Three parts, and the middle one is the reason this is not a two-line change:

1. **`save()` writes NULL, not `''`, for an empty optional date/time field.**
   The column types are already in `commons/schema-expected.php`, which is what
   `OpenAPI::_entitySchema()` reads for exactly this kind of question, so the
   ORM can find them without a round trip.
2. **A schema step making the 13 `NOT NULL` columns nullable, and repairing
   existing zero dates to NULL.** Without this, part 1 is a no-op at best: an
   explicit NULL into a `NOT NULL` column is error 1048 under strict mode and,
   with `sql_mode` cleared, is coerced right back to `0000-00-00`. Step 343
   (#1244) already did this for one column and is the template.
3. **Then, and only then, the `SET SESSION sql_mode=''` can be narrowed or
   dropped** — as its own change, with its own testing. Not in the same commit.

Deliberately **not** proposed: leaving `sql_mode` cleared and calling it
supported. Note that option — "set a session `sql_mode` to paper over the write"
— is what the code does today, in its strongest form, for every statement FOG
issues.

Open question for part 2: what reads these columns. `0000-00-00` and NULL format
differently, sort differently, and compare differently, so the display and
comparison sites need a sweep before the ALTER lands. That sweep is the next
step, and it decides whether this is one PR or two.

mastacontrola:
Second of two for #1245. **Stacked on #1247** — based on that branch, so merge
#1247 first. Everything here moves data; #1247 is the display half that makes
that safe.

## The write

`FOGController::save()` writes `''` for any unset optional field whose key does
not end in `id`. A date column cannot hold `''`: the server refuses it under a
strict sql_mode and coerces it to `'0000-00-00 00:00:00'` without one. FOG only
ever sees the second, because `PDODB::_connect()` issues `SET SESSION
sql_mode=''` on every connection.

That is why this server — MariaDB 11.8 with `STRICT_TRANS_TABLES` **in its own
config** — holds 83 zero `hostLastDeploy` out of 86 hosts and 85 zero
`hostSecTime`. Values its configuration forbids, stored because FOG asked it to
stop checking.

## Three parts

**1. `save()` decides from the column's TYPE, not the key's name** — the lesson
`$databaseFieldsNotInt` already exists for. `commons/schema-expected.php`
carries per-column types and is what `OpenAPI::_entitySchema()` reads for the
same question, so no new source of truth; a missing manifest degrades to
today's behavior.

The value is bound as a **real NULL, not dropped from the statement**. `ON
DUPLICATE KEY UPDATE` only touches columns the statement names, so omitting it
would mean an existing date could never be cleared — the write would report
success and change nothing.

**2. Schema step 344** makes eleven columns nullable and repairs the rows
already holding a zero date. Without it part 1 is worse than a no-op: an
explicit NULL into a `NOT NULL` column errors under a strict mode and is
coerced straight back to the zero date without one. ALTER *before* UPDATE — the
opposite order to step 343, because the rows cannot be set NULL until the
column can hold NULL.

Eleven is every date column that is optional, not auto-filled, and without a
server-side default. Two reachable columns are deliberately left `NOT NULL`:
`snapinTasks.stCheckinDate` and `userAuths.uaExpireDate` both declare `DEFAULT
current_timestamp()`, and `UserAuth::reapExpired()` deletes on `uaExpireDate <
now` — NULL satisfies no comparison, so a nullable expiry would turn a token
that fails safe into one that is never reaped at all.

**3. The four sites that used the zero date directly.** Two of them are **query
filters**, not writes — `TaskingElement::imageLog()` finding the imaging log
that has not finished.

## Please look at this one

`FOGManagerController::distinct()` now reads a **null filter as `IS NULL`**.
Bound as a placeholder it was `col = NULL`, which is never true, so those two
lookups would have silently found nothing once the column went NULL. It is a
shared query builder, which is why I am flagging it rather than burying it: the
change is additive — a null filter could not previously match anything, so no
working behavior can change — but it is the one edit here outside the date
story.

`perform_update()` likewise stops running `trim()` over a null update value,
which turned it back into `''` (and was a PHP 8.1 deprecation).

## Proof

Booting the actual stack — `commons/base.inc.php`, the autoloader, PDODB,
FOGController — against a throwaway server carrying a freshly replayed
`schema.php`. `scripts/background_scripts/prove_save_writes_null_1245.sh`:

```
== BEFORE: save()'s date arm removed ==
  hostLastDeploy   '0000-00-00 00:00:00'
  set a real date  '2026-01-02 03:04:05'
  clear it again   '0000-00-00 00:00:00'   <- an existing date could not be cleared
  string hostIP    ''

== AFTER: as shipped ==
  hostLastDeploy   NULL
  set a real date  '2026-01-02 03:04:05'
  clear it again   NULL
  string hostIP    ''
```

Note the column was already nullable in both runs — FOG *chose* the zero date
where NULL was available, which is the arm, not the schema.

`verify_null_dates_1245.sh` proves step 344 repairs an install that already
carries zero dates: it builds the pre-344 shape by hand, inserts `''` the way
FOG does, runs the step's own array literal, and reads back
`information_schema`. All eleven columns `nullable=YES value=NULL`.

`tests/date-columns-nullable.test.php` holds the invariant, which spans a model
and the schema and cannot be checked in either alone: a date column whose model
field is optional and not auto-filled must be nullable or carry a server
default. Both exemptions are listed with their reason, and an exemption that
stops being true fails too. Three mutations, all killed. Full suite green
locally.

## Manifest

`schema-expected.php` is hand-edited for exactly the eleven columns, in both
its `create` and `columns` blocks. I regenerated it properly first, on 11.8,
and backed that out — the generator produced three unrelated changes that do
not belong in this PR, one of which is a regression. Separate comment below.

## Still not in scope

The `SET SESSION sql_mode=''` itself. It comes off once the data is clean, on
its own, with its own testing.

Closes #1245


mastacontrola:
## Why the manifest is hand-edited here, and not regenerated

The header says the `tables` block is generated and asks nobody to hand-edit
it, so this needs justifying rather than just doing.

I did regenerate it properly first —
`scripts/background_scripts/regen_schema_manifest_1245.sh`, which starts a
throwaway server, replays `commons/schema.php` into it, and runs
`bin/schema-manifest.php generate` against the result. Two things came out of
that, and only one of them was mine.

**The server version matters, and it is not recorded anywhere.** Generated on
`mariadb:10.5`, every `create` string in the file changed: 10.5's
`SHOW CREATE TABLE` says `CHARSET=utf8 COLLATE=utf8_general_ci` where 11.x says
`utf8mb3` / `utf8mb3_general_ci`. The committed file says `utf8mb3`, so it was
generated on 11.x. Regenerating on the wrong one silently rewrites all 67
tables and, on a strict reading, un-pins the collation that #1152 just pinned.
Nothing in the file or the generator says which version produced it.

**A fresh-replay regeneration is not equivalent to the committed file.** Three
differences, none of them mine:

| difference | verdict |
|---|---|
| `aloLog` added | harness artifact — the table is dropped at step 238 by a `Schema::dropTable()` **closure**, and the replay skips closure steps ("11 closure step(s) skipped"). The manifest is right to omit it. |
| `taskLog` loses `logType`, `logText`, `logHostID`, `logHostName`, `logTaskTypeName`, and `taskID` becomes `mediumtext` | same cause — those come from the reconcile step, which is also a closure. The manifest is right. |
| `ngmInterface` default changes from `'enp58s0u2u4'` to `''` | **a real defect in the shipped file.** Filed as #1249. |

That third one is worth reading: step 26 builds the column with
`DEFAULT '<STORAGE_INTERFACE>'`, which is per-install, so the manifest captured
one machine's NIC name as the default for everyone. `SchemaReconciler` uses that
text verbatim when it has to add the column.

So a wholesale regeneration would have shipped two regressions and an unrelated
fix inside a date-columns PR. The eleven columns are edited in both the `create`
and `columns` blocks instead, and
`tests/schema-manifest-consistent.test.php` (647 checks) confirms the two blocks
still agree.

The regeneration script is kept because it is the right tool once #1249 is fixed
and the closure-step gap is understood — it just needs `IMAGE=mariadb:11.8`,
which it now defaults to.

PR #1258, opened 2026-08-21
fix(core): report and record database operations that fail on machine paths (1.5 port of #1257)

mastacontrola:
Port of #1257 (working-1.6). Same two defects, same fix. **Merge #1257 first** — this is the backport, not the other way round.

## The problem

A failed database operation on a machine-facing endpoint was recorded nowhere, and on the commonest cases it was reported as a **success**.

1. `PDODB::query()` catches the `PDOException` and returns normally (`$throwOnQueryError` defaults false), so `save()` never entered its own catch on a real SQL error.
2. For an existing row there was no `insertId()` check to trip, so `save()` logged "successfully updated" and returned `$this`.
3. `logHistory()` returns early unless `self::$FOGUser` is a valid `User`, and nothing on a machine path has one — `service/`, `lib/reg-task/` and the daemons are matched to a **host** by MAC or token.

**This branch is worse than 1.6 at step 3, which is why the port is not optional.** 1.6's `debug()` at least reaches a file, behind a globalSetting that ships off. This one writes to no file **at all** — it `printf()`s into the page and returns immediately when `self::$service` or `self::$ajax` is set, which on a machine endpoint is always. A failed write on a service path had *literally no possible output*.

## The fix

**Part A** — `save()`, `destroy()` and `load()` read `self::$DB->error` immediately after their statement. Not `PDODB::$throwOnQueryError = true` globally, which would turn every already-tolerated failure in the codebase into an uncaught 500 at once.

**Part B** — new `FOGBase::logFault()`: one line under FOG's log directory, `error_log()` as the fallback. Deliberately **not** `logHistory()` with the gate widened: `logHistory()` is a database write, so it shares its failure mode with the thing it reports on. **A sink for a failed database operation cannot itself be a database write.** It reads no setting and issues no query — a logger that queries in order to report a failed query is a recursion this project has already paid for.

Two files rather than one, split by SAPI. This is the only FOG log directory written by **both** tiers, and a shared file would be owned by whichever wrote first — a root-owned `faults.log` appears the moment a daemon hits a failed write, and every web-tier fault silently falls back from then on.

**The read path** has only the second defect and is quieter for it. A swallowed SELECT hands back nothing, so an object that could not be **read** is indistinguishable from a row that genuinely holds no data. There is no return value to get wrong — `load()` answers `$this` either way, and must, or `new Host(42)` becomes fatal on a database blip — so the log line is the entire fix. It is recorded at the error check rather than in `load()`'s catch, because that catch also handles the method's *ordinary control flow*: "Operation field not set" fires on every object built without an id, and at that volume it would bury the one line worth having.

`FOGManagerController` had it in **seven** places — `update()` (`(bool)` of an object, always true), `destroy()` (returned `true` unconditionally), `uninstall()` (returned the PDODB object while documented `@return bool`), `insertBatch()`, and the reads `find()`, `exists()`, `count()`/`distinct()`. `exists()` gives the most expensive wrong answer in the class: callers use it to decide whether to **create**, so an unreadable database becomes a duplicate row rather than an error. Read return contracts are left alone — making a read throw is a real change that belongs in its own decision.

`TaskError::_logRow()` is covered **with no special case**, which is why this went in the framework: it discards `save()`'s return *by design*, because it **is** the sink FOS reports land in.

## Differences from #1257, all forced by this branch

| | |
|---|---|
| No `FOG_LOG_DIR` constant exists here | `FAULT_LOG_DIR` spells `/opt/fog/log` out the way `TaskError::LOG_DIR` already does, for the reason its docblock gives. `FOG_LOG_DIR` is still preferred when defined — costs nothing, keeps both branches the same shape. |
| No `FOGLogPaths` here | `faults/` added to all **three** hardcoded lists (`getfiles.php`, `logtoview.php`, `StorageNode`) — the duplication `FOGLogPaths` was created to end. |
| No `loadMany()`, no per-plugin-table column lookup | `_loadColumnTypes()` is the whole-catalog equivalent and gets the same treatment. |
| `FOGManagerController::destroy()` and `find()` exist only here | Both covered. |

## Behaviour change — deliberate

`save()` and `destroy()` now return `false` where they returned the object, and the manager's `update()`, `destroy()` and `uninstall()` return a real bool. Callers that check the return start seeing failures that were previously silent. That is the fix. The read paths' return contracts are unchanged.

## Verification

- `tests/db-failure-is-recorded.test.php` is **behavioural, not source-reading** — it reuses `tests/lib/scope-harness.php`, which boots FOG against a fake database. That was better than expected for this branch.
- **Six mutations caught**, including the one that moves `load()`'s fault into the catch, which proves the control-flow separation is real rather than decorative.
- **Suite 21/21.**
- **The seam checked against a real server using this branch's own PDODB:** a valid statement leaves `->error` falsy, an invalid one sets it, and a valid one after a failure clears it again — so one bad query cannot fault every statement after it.

**What was not done here, stated plainly:** the full end-to-end live run (a real `save()` rolled back, a real `load()`, `exists()`, `distinct()`) was performed on working-1.6, not on 1.5. The only 1.5 servers in the lab are remote with no sudo, so their database is not reachable from here, and running this branch's ORM against the local 1.6 database would be asserting against a schema it was not written for. `PDODB::query()` is materially identical on both branches — both set `$this->error = false` at the end of the try — which is what makes the 1.6 run transfer.

### DIFF
--- a/lib/common/functions.sh
+++ b/lib/common/functions.sh
@@ -902,6 +902,21 @@ installFOGServices() {
     # without this the directory exists, looks right, and every report is
     # dropped with nothing but an AVC to say so.
     setSELinuxContext "$servicelogs/fos" httpd_sys_rw_content_t
+    # Where FOGBase::logFault() records database operations that did not
+    # happen. Its own subdirectory for the same reason the one above has its.
+    #
+    # Unlike that one, BOTH tiers write here -- the web user, and root for the
+    # daemons -- so logFault() writes faults-web.log and faults-service.log
+    # rather than one shared file, whose owner would be whichever tier hit a
+    # failure first. The directory is the web user's; root writes into it
+    # regardless of mode.
+    dots "Creating FOG fault log directory"
+    mkdir -p $servicelogs/faults >>$error_log 2>&1
+    chown ${apacheuser}:${apacheuser} $servicelogs/faults >>$error_log 2>&1
+    errorStat $?
+    # Outside the dots/errorStat pair, like every other caller, and the _rw_
+    # label is as load-bearing here as it is for fos above (GH-964).
+    setSELinuxContext "$servicelogs/faults" httpd_sys_rw_content_t
 }
 configureUDPCast() {
     dots "Setting up UDPCast"
--- a/packages/web/lib/db/pdodb.class.php
+++ b/packages/web/lib/db/pdodb.class.php
@@ -556,6 +556,28 @@ class PDODB extends DatabaseManager
                 $this->sqlerror()
             );
             self::$_result = false;
+            /*
+             * $msg used to be built here and dropped on the floor: a failed
+             * fetch set no ->error, logged nothing, and left $_result false
+             * -- so get() answered an empty set and the caller could not tell
+             * "the read failed" from "there are no rows". That is the same
+             * defect FOGController::save() and load() were carrying, one
+             * layer down, and it is why those checks alone were not enough.
+             *
+             * Only ever ADDS a failure, never overwrites one. query() owns
+             * clearing ->error -- it runs immediately before every fetch()
+             * and always sets it to false or to a message -- so when a fetch
+             * fails BECAUSE the query did ("No query result, use query()
+             * first"), the guard keeps the original cause rather than
+             * replacing it with the symptom.
+             *
+             * Not logged from here. The callers know which class and table
+             * they were reading, and this does not; a line naming neither is
+             * worse than the caller's, and two lines per failure is noise.
+             */
+            if (!$this->error) {
+                $this->error = $msg;
+            }
 
             if (self::$throwOnQueryError) {
                 throw $e;
--- a/packages/web/lib/fog/fogbase.class.php
+++ b/packages/web/lib/fog/fogbase.class.php
@@ -768,6 +768,175 @@ abstract class FOGBase
         printf('<div class="debug debug-error">%s</div>', $string);
     }
     /**
+     * FOG's log directory.
+     *
+     * 1.5 has no FOG_LOG_DIR constant -- every service spells this path out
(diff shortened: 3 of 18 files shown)


mastacontrola:
Review pass found a defect the first commit's checks would have missed, fixed in eb3c705.

`PDODB::fetch()` built its error message in the catch and **dropped it on the floor** — `$msg` assigned, never used. A fetch failure set no `->error`, logged nothing, and left `$_result` false, so `get()` answered an empty set with nothing to distinguish it from "there are no rows". Same defect as the one this PR fixes in `save()`/`load()`, one layer down — and the reason the first commit's checks were not enough: they sat *between* `query()` and `fetch()`, covering a rejected statement and missing a rejected read of its rows.

`fetch()` now records onto `->error`, and only when nothing is there already. `query()` owns clearing it and runs immediately before every fetch, so when a fetch fails *because* the query did, the guard keeps the original cause rather than replacing it with "No query result, use query() first". Not logged from inside PDODB — the callers know which class and table they were reading and it does not.

With fetch recording and never clearing, **one** check after the fetch covers both halves, so `load()`, `exists()`, `find()` and the two count-shaped methods moved their check there. Fewer checks than checking twice, and it reads the right message either way.

Worth flagging: **the first version of this gate did not hold.** The test's fake `fetch()` modelled the fixed behaviour, so gutting the real `PDODB::fetch()` still passed — pinning the symbol's use rather than its definition, which is precisely how a gate survives the mutation it exists to catch. The assertion now drives the real method through reflection. Reverting either half of the fix fails the suite.

Also present on `working-1.6` (identical dead `$msg`); #1257 is merged, so that goes in as a follow-up PR rather than an amend.

CI 7/7, suite 21/21.

PR #1269, opened 2026-08-21
Give the installer one absolute scratch directory, and stop hiding why it failed

darksidemilk:
Give the installer one absolute scratch directory, and stop hiding why it failed

Reported twice from the field:

    Downloading kernel, init and fog-client binaries..............
    lib/common/functions.sh: line 9059: cd: ../tmp/: No such file or directory

Five call sites wrote the same three lines:

    [[ ! -d ../tmp/ ]] && mkdir -p ../tmp/ >/dev/null 2>&1
    cd ../tmp/

Three faults, and they compound into something worse than the message suggests.

  * ../tmp/ resolves against whatever the ambient cwd happens to be, so the
    answer moves depending on who called and whether anything cd'd earlier.
  * The mkdir error goes to /dev/null. The one message that would say WHY is
    destroyed at the instant it is produced, which is why two reports of this
    contain no diagnosis between them.
  * The cd is unguarded, so execution CONTINUES in the wrong directory.

That last one is the expensive part. copypath was the empty string, so every
copy after the download loop was relative too -- meaning a failed cd downloaded
60-80MB of kernels, inits and installers into bin/, and then copied them out of
bin/ successfully. The install appeared to work. It left untracked binaries in
the source tree (.gitignore covers tmp/* and *.exe, not bzImage, init.xz or
*.msi) and the retry loop found them again on the next run, which is what made
this look transient the first time it was reported.

It is not transient. The second report did not clear on a re-run, and that is
the signature of the path existing as a non-directory: mkdir -p refuses a file
or a dangling symlink every time, so every retry fails identically. The generic
wording sent the reporter looking for a permissions problem they did not have.

_installerTmpDir() now owns the answer. It anchors on $workingdir, which
installfog.sh captures with pwd before anything can move, so it is correct
regardless of the caller's cwd; it returns an absolute path; it reports the
not-a-directory case by name and says to remove or rename the path; and it fails
non-zero so callers can stop.

All five call sites converted:

  downloadfiles()          fails the install rather than downloading into bin/.
                           copypath is set to the scratch directory as well, so
                           the copy step no longer depends on cwd either -- that
                           variable was already the hook for it and had only ever
                           held "".
  configureUDPCast()       same, plus a guard on `cd $udpcastout`. Without it a
                           tarball that did not unpack ran configure, make and
                           make install in the scratch directory instead, each
                           failing separately and none of them saying why.
  fetchipxeasset()         was the least broken -- it logged its mkdir error --
                           but still resolved ../tmp/ against the cwd.
  the two SQL setup paths  no cd, so these failed differently: the heredoc wrote
                           into a missing directory and mysql was then handed a
                           file that had never been created. Same cause, same
                           discarded error.

cwd/cur are now local and quoted at those sites; cwd in downloadfiles had been a
global.

tests/installer-tmpdir.test.sh pins it: the path is absolute, it does not move
with the caller's cwd, creation is idempotent, a path that exists as a file fails
with a message naming THAT cause and offering a remedy, failure is visible in the
exit status, and no call site has quietly gone back to the relative form.
Mutation-verified -- returning a relative path fails 7 assertions, and removing
the not-a-directory guard fails the two that check the message is specific rather
than generic. 11 passed, 0 failed. Picked up automatically by run-all.sh's glob.

What this does NOT do is explain the original failure. The reason was thrown away
by the code being fixed here. If it recurs, the message will now name the cause.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/lib/common/functions.sh
+++ b/lib/common/functions.sh
@@ -1162,8 +1162,17 @@ updateDB() {
             fi
         fi
     fi
-    [[ ! -d ../tmp/ ]] && mkdir -p ../tmp/ >/dev/null 2>&1
-    cat >../tmp/fog-db-grant-fogstorage-access.sql <<EOF
+    # Same scratch directory as everything else. This one does not cd, so the
+    # old code failed differently -- the heredoc write failed, then mysql was
+    # handed a missing file -- but from the same cause, and with the mkdir error
+    # discarded just the same.
+    local sqltmp=""
+    if ! sqltmp="$(_installerTmpDir)"; then
+        echo "Failed"
+        echo " * Could not prepare the installer's scratch directory."
+        return 1
+    fi
+    cat >"${sqltmp}/fog-db-grant-fogstorage-access.sql" <<EOF
 SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE='ANSI' ;
 GRANT SELECT ON $mysqldbname.* TO 'fogstorage'@'%' ;
 GRANT INSERT,UPDATE ON $mysqldbname.hosts TO 'fogstorage'@'%' ;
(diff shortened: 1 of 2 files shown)


darksidemilk:
merged this and then the log made it clear...I was just out of space...

PR #1285, opened 2026-08-22
Describe join, the hostname rule, and upsert accurately in the OpenAPI document

darksidemilk:
Three corrections to `OpenAPI::document()`. One file, `packages/web/lib/fog/openapi.class.php`, +226/-8. No route behaviour changes.

## What was wrong

**`/{class}/join` was described as one thing when it is two.** `Route::joining()` serves two different operations on that path and the document collapsed them:

- `PUT` is a **bulk edit** over an `ids` array, answering `202` with no body
- `POST` is **group-only**, a get-or-create by name answering `201`

Added `_bulkEditBody()`, `_namesBody()`, `_acceptedResponse()` and `_idsResponse()`, plus an `$operationId` override on `_op()` so the two operations on one path get distinct ids.

**The hostname rule was invisible.** `host.name` is constrained server side to 1–15 characters from a fixed set, and a caller only found out by getting a rejection back. The column type cannot express it, so `_applyModelConstraint()` now adds `maxLength: 15` and the pattern to that one field. Callers that validate against the document now fail locally instead of at the server.

**"upsert" was used without being defined.** The join prose leaned on the term; it now says what it means inline.

## Why no route change

Nothing here touches `route.class.php`. These are description-only corrections to operations the router already served this way — the document was under-describing behaviour that already existed.

## Downstream

FogApi (`darksidemilk/FogApi`) reads this document to generate its cmdlets. Its snapshot has been regenerated from this branch and verified byte-identical to what a server built from it serves, apart from the deliberate `servers[0].url` placeholder.

## Separately found, not fixed here

While generating against this document, `/{class}/search/{item}` turned out to be undeliverable for classes with no `name` column. `Route::search()` does not query the class — it calls `unisearch()` and reads the class out of the result — and `unisearch` deliberately skips entities with no `name` field. Measured against a live server on `tasklog` with `42`, `probe`, `Win-11-23H2` and `Deploy`, all of which matched real rows: every one returned `recordsFiltered: 0` against `recordsTotal: 4`.

`_paths()` emits the generic ten operations for every route class, so **20 entities are documented with a search route they cannot honour**: `filedeletequeue`, `groupassociation`, `history`, `hostautologout`, `hostscreensetting`, `imageassociation`, `inventory`, `ipxe`, `macaddressassociation`, `moduleassociation`, `multicastsessionassociation`, `nodefailure`, `powermanagement`, `printerassociation`, `snapinassociation`, `snapingroupassociation`, `snapinjob`, `snapintask`, `tasklog`, `usertracking`.

The fix is to emit `search` only where the model has a `name` field — the same condition `unisearch` already applies. Left out of this PR to keep it to the three described corrections; happy to add it here or raise it separately.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VhYS2cgaUEgEBuFfsPfU5B)_


darksidemilk:
Following up on the "Separately found, not fixed here" section: the search fix is done, on `claude/openapi-search-named-entities-xk2ybx` (branched from `working-1.6`, commit `4350d50`).

`_isSearchable()` applies the same `isset($vars['databaseFields']['name'])` that `unisearch()` applies, against the reflected metadata `_classVars()` already caches — not a list of class names, since `$validClasses` is mutated at runtime by `API_VALID_CLASSES` and a hand-kept list would silently mis-describe every class a plugin adds. `unisearch()` and `Route::search()` are untouched; this is description-only, exactly as this PR was.

### Verified against a live server

Built a 1.6.0-beta.3840 server from `working-1.6` via `bin/installfog.sh` and measured before and after.

Reproducing the original finding first — four seeded `taskLog` rows, all four terms matching real ones:

| `/tasklog/search/{item}` | recordsFiltered | recordsTotal |
|---|---|---|
| `42` | 0 | 4 |
| `probe` | 0 | 4 |
| `Win-11-23H2` | 0 | 4 |
| `Deploy` | 0 | 4 |

Control on a class that does have a name: `/host/search/LAB` → `recordsFiltered: 1` for a host that exists.

After the patch, against `/system/openapi` on the same server:

- search paths **51 → 31**, removing exactly the 20 listed here and nothing else
- no other path added or lost; tags 52 → 52, schemas 53 → 53
- no tag orphaned (every class still emits list/indiv/count/names/ids)
- 525 operations, no duplicate `operationId`, no dangling `$ref`
- the 20 routes answer exactly as before — no route behaviour changed

Also corrected on the way past: the `item` parameter said the term is matched "across the class fields", which was never true. `unisearch()` matches id and name, plus one extra arm for `host` (MAC), `storagenode` (node hostname) and `setting` (value). It now says that.

### Separately found, not fixed here

The name-field test is necessary but not sufficient, so this does not make every remaining search route honest.

`Route::search()` reads `$items->{$classname}` out of the `unisearch()` payload, and `unisearch()` iterates `FOGBase::$searchPages` — 16 entries — not `$validClasses` (51). It also `continue`s on `task` at the top of the loop. So only 15 classes can ever be populated: `group`, `host`, `image`, `ipxe`, `module`, `plugin`, `printer`, `role`, `setting`, `site`, `snapin`, `storagegroup`, `storagenode`, `user`, `usergroup`. The other 16 that survive this patch — `hookevent`, `imagepartitiontype`, `imagetype`, `keysequence`, `multicastsession`, `notifyevent`, `os`, `oui`, `pxemenuoptions`, `rolepermission`, `roleuserassociation`, `roleusergroupassociation`, `scheduledtask`, `task`, `taskstate`, `tasktype`, `usergroupmember` — have a name column but are not in `$searchPages`, so they answer empty for the same reason `tasklog` did.

Narrowing to that is a larger claim about what the document promises, and it would change which cmdlets FogApi retires downstream, so I left it out rather than widen this on my own.

One genuine bug rather than a documentation gap, worth its own issue: `unisearch()` remaps `ipxe` → `pxemenuoptions` for the model lookup but keys results under `ipxe`. `Route::search('ipxe', …)` then feeds **pxeMenu ids into `ipxeTable`** via `listem('ipxe', ['id' => $ids])`, so it returns whatever `ipxeTable` rows happen to share those ids — wrong rows, not zero rows. This patch stops documenting `/ipxe/search/{item}` (the `Ipxe` model has no `name`), but the route still behaves that way.

Unrelated, found while installing: `bin/installfog.sh` fails at "Granting access to fogstorage database user" with `ERROR 1146: Table 'fog.imagingLog' doesn't exist` — the grant still names a table 1.6 has retired.

---

PR #1328, opened 2026-08-23
feat(api): hashed, revocable API tokens as the only Bearer credential

mastacontrola:
Implements [ADR 0027](../blob/feat-api-token-store/docs/adr/0027-api-tokens-are-a-separate-hashed-credential.md), merged in #1327.

## What this adds

A new `apiTokens` table and `APIToken` model, and `Authorization: Bearer` repointed at it. A Bearer credential is now a distinct thing from `users.uAPIToken` rather than a second spelling of it:

| Property | `uAPIToken` (fog-user-token) | `APIToken` (Bearer) |
|---|---|---|
| At rest | plaintext | SHA-256 |
| Visible in the UI | yes, any time | once, at creation |
| Per user | exactly one | as many as you like |
| Revocable individually | no | yes, per-token enabled flag |
| Independent of `uAllowAPI` | n/a | yes |
| Needs `fog-api-token` alongside | yes | no |

Tokens carry a `fog_` prefix so they are recognisable in a log or a paste, and so leaked-credential scanners have something to match.

## What does not change

`users.uAPIToken` is **untouched**. It stays plaintext, stays visible in the UI, and keeps working as `fog-user-token` beside `fog-api-token`. Nothing has to migrate, and no existing client that uses the legacy pair is affected.

## Behaviour change for one interim case

#1324 accepted `uAPIToken` as a standalone Bearer value. **That is withdrawn.** A plaintext, UI-visible, non-revocable secret is exactly what a Bearer credential must not be, so accepting it there gave the scheme none of the properties in the table above. The legacy two-header pair remains the supported path for that token.

This is worth flagging explicitly for anyone who adopted Bearer between #1324 and this PR: swap to a token issued from the API tab, or fall back to the header pair.

## Design notes

- **Unsalted SHA-256, deliberately.** A salt defeats precomputation, which needs a guessable input; these are 512 bits of CSPRNG output, so a salt protects nothing while costing the ability to look a token up at all. The full reasoning, and the invariant it rests on, are in `APIToken::hashToken()`'s docblock. If a token is ever shortened or made user-choosable that reasoning is void — the test enforces the entropy floor so that gets noticed.
- **`APIToken` is not in `Route::$validClasses`.** A token-management REST surface lets one leaked credential mint another, so revoking the one that leaked would fix nothing. Creation and revocation are UI actions, behind a session and CSRF.
- **`userAPIPost()` routes to the token handler first and returns.** Both cards post to the same tab URL; falling through would read the legacy card's absent fields as "unticked" and "empty", silently disabling `fog-user-token` and wiping `uAPIToken` as a side effect of issuing a Bearer token. That is the GH-987 control-type defect class exactly.
- **`resolve()` fails closed on an ownerless token** as well as cascading on user delete. The cascade is the fix; the fail-closed is what makes a mistake in it survivable.

## Verification

Against the local 1.6 lab, over real HTTP and through the model. Probes are committed at `scripts/background_scripts/probe_api_token_store.sh` and `probe_api_token_lifecycle.php` in the maintainer's script tree.

```
== 1. new APIToken, standalone Bearer, raw ==
  ok    GET /host with Bearer fog_...                              200
  ok    GET /user/1 with Bearer fog_...                            200
== 2. legacy pair still works, untouched ==
  ok    GET /host with fog-api-token + fog-user-token              200
== 3. uAPIToken is no longer a Bearer credential ==
  ok    Bearer <uAPIToken raw>                                     401
  ok    Bearer <uAPIToken base64>                                  401
== 4. malformed and absent Bearer ==
  ok    no Authorization header                                    403
  ok    bare 'Bearer' with no value                                401
  ok    'Bearertoken' (no delimiter)                               403
  ok    garbage Bearer value                                       401
== 5. secrets are not echoed back ==
  ok    GET /user/1 omits 'token'                                  absent
  ok    GET /user/1 omits 'password'                               absent

passed 11, failed 0
```

```
== A. per-token enabled flag is independent of uAllowAPI ==
  ok    enabled token resolves                                   117
  ok    disabled token refused                                   NULL
  ok    ...while the user is still API-enabled                   '1'
  ok    re-enabled token resolves again                          117
== B. lastUsed is stamped, and never as an empty date ==
  ok    atLastUsed is set                                        true
  ok    atLastUsed is a real datetime, not an empty string       true
== C. cascade on user delete ==
  ok    token exists before delete                               1
  ok    tokens removed with their owner                          0
  ok    orphaned token fails closed                              NULL

passed 9, failed 0
```

`tests/api-token-store.test.php` pins the same contract in CI, where no database is available — 39 checks. Mutation-verified: reintroducing `base64_decode`, dropping the cascade entry, removing either guard in `resolve()`, dropping `UNIQUE` from `atHash`, and shortening the token are all caught.

Full suite: **136 passed, 0 failed** on PHP 8.3.33.

## Schema

Adds element `// 359` (`apiTokens`) and bumps `FOG_SCHEMA` to match, with the manifest entry in `schema-expected.php`.

## OpenAPI

No REST route changed, so `OpenAPI::document()` still describes the API correctly. The `bearerAuth` scheme it already advertises (added in #1324) is now backed by this store rather than by `uAPIToken`.

## Downstream

None. `Route::$validClasses` is unchanged, so FogApi's hardcoded class list needs no sync.

## Follow-up

`docs/kb/integrations/api.md` in fog-docs describes the Bearer token as `uAPIToken` sent raw. That needs updating to the new format and the issue-from-the-UI flow — I'll open that PR separately so this one isn't gated on it.

https://claude.ai/code/session_01XqpPXAk7bEm8huH9WkiGk6

### DIFF
--- a/packages/web/commons/schema-expected.php
+++ b/packages/web/commons/schema-expected.php
@@ -88,6 +88,19 @@ return [
                 'acRedacted' => 'tinyint(1) unsigned NOT NULL DEFAULT 0',
             ],
         ],
+        'apiTokens' => [
+            'create' => 'CREATE TABLE IF NOT EXISTS `apiTokens` ( `atID` int(11) NOT NULL AUTO_INCREMENT, `atUserID` int(11) NOT NULL DEFAULT 0, `atName` varchar(255) NOT NULL DEFAULT \'\', `atHash` char(64) NOT NULL DEFAULT \'\', `atEnabled` enum(\'0\',\'1\') NOT NULL DEFAULT \'1\', `atCreatedTime` datetime NOT NULL DEFAULT current_timestamp(), `atCreatedBy` varchar(255) NOT NULL DEFAULT \'\', `atLastUsed` datetime DEFAULT NULL, PRIMARY KEY (`atID`), UNIQUE KEY `atHash` (`atHash`), KEY `atUserID` (`atUserID`) ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_general_ci ROW_FORMAT=DYNAMIC',
+            'columns' => [
+                'atID' => 'int(11) NOT NULL',
+                'atUserID' => 'int(11) NOT NULL DEFAULT 0',
+                'atName' => 'varchar(255) NOT NULL DEFAULT \'\'',
+                'atHash' => 'char(64) NOT NULL DEFAULT \'\'',
+                'atEnabled' => 'enum(\'0\',\'1\') NOT NULL DEFAULT \'1\'',
+                'atCreatedTime' => 'datetime NOT NULL DEFAULT current_timestamp()',
+                'atCreatedBy' => 'varchar(255) NOT NULL DEFAULT \'\'',
+                'atLastUsed' => 'datetime DEFAULT NULL',
+            ],
+        ],
         'auditLog' => [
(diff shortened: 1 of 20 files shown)


mastacontrola:
## Verified on a real install

Deployed to a 1.6 lab server, ran the schema updater, and drove the UI in a browser. That found **four defects no source read would have** — every one silent, with the page rendering and the buttons doing nothing. All fixed in bb8d64e7a; details in that commit message. The short version:

- `APITokenManager::find()` did not exist. There is no generic "rows matching X" on `FOGManagerController`, so the user edit page fataled with a bodyless 500. Replaced with `forUser()`.
- The card had **no JS wiring at all**. FOG has no generic binding that picks a tab form up — every card in `fog.user.edit.js` is wired by hand.
- The create/save discriminator could never have arrived: `processForm()` posts `new FormData(form)`, and FormData omits submit buttons. Saving now keys on a hidden field; issuing is its own CSRF-gated endpoint, which also fixes show-once properly.
- The endpoint sent JSON with no `Content-type`.

That last one exposed a **latent bug in `fog.common.js`** worth flagging separately: `notifyFromAPI` never declared `msg`. Every branch assigned it as an implicit global and `if (!msg)` *read* it — so any response carrying none of `error`/`info`/`warning`/`msg` threw `ReferenceError` out of the success handler and took the caller's callback with it. The `'Bad Response'` fallback two lines below existed for exactly that case and could never run. Declared it.

### Browser walkthrough

| Step | Result |
|---|---|
| Issue a token from the API tab | show-once banner appears with the plaintext |
| `Authorization: Bearer fog_…` (raw) against `/fog/host` | **200** |
| Same token base64-encoded | **401** |
| `atLastUsed` / `atCreatedBy` after that call | `2026-08-23 16:11:05` / `fog` |
| Reload the tab | row listed, banner gone |
| Untick **Enabled**, Update | token now **401** |
| Tick **Delete**, Update | row gone |

Throughout all of it, `uAllowAPI` stayed `1` and `uAPIToken` stayed intact — the GH-987 clobber the routing guard exists to prevent does not fire.

The schema step applied cleanly through the normal updater: `vValue` 358 → 359, `apiTokens` created.

`tests/api-token-store.test.php` gains eight checks covering the wiring, the FormData trap and show-once, so these silent shapes are pinned rather than rediscovered. 47 checks; suite 136 passed, 0 failed; CI green on both PHP versions and all three database engines.

PR #1351, opened 2026-08-25
Remove dead code in BootMenu and pin its iPXE output with a test

mastacontrola:
## What this changes

Five write-only or no-op statements in `bootmenu.class.php`, plus one variable that was never defined at all.

| Site | Problem |
|---|---|
| `__construct()` | `$reboot = sprintf('reboot', "\n")` — format string with no placeholder, argument with nowhere to go |
| `__construct()` | `$kernel = $bzImage` assigned, never read; `$this->_kernel` is built from `$bzImage` directly |
| `_chainBoot()` | `$debug = $debug;` — self-assignment left over from a signature change |
| `printDefault()` | `PXEMenuOptionsManager->find('', '', 'id')` result assigned then overwritten by the `pxeRegOnly`-filtered query below — one wasted query per menu render |
| `__construct()` | `BOOT_ITEM_NEW_SETTINGS` passed `'webroot'` by reference with no `$webroot` ever assigned, so PHP created it at the call site and every plugin reading it received `NULL` |

The first four are pure removals. The `webroot` one is a real fix: the argument is now bound to the bare webroot, the same value `set fog-webroot` emits, so a plugin reading it gets what the name promises. Nested webroots (`/apps/fog/`) keep every segment.

Net: 18 lines changed in the class, 7 of them deletions.

## How it is proven

The whole claim for the removals is *"not one byte of emitted iPXE moves."* That is not something inspection can settle on a 2400-line file whose output is a program, so this adds a test that asserts it directly.

`tests/bootmenu-ipxe-output.test.php` renders 37 scenarios through a stub FOG runtime and byte-compares the emitted script against a golden file.

**The golden was generated from this file as it stood before the removals.** A passing run is therefore the proof that they were no-ops, not merely an assertion that they were.

```
bootmenu-ipxe-output: 37 scenarios, 101211 bytes rendered
bootmenu-ipxe-output: 2 hook-argument checks passed
PASS
```

Regenerating the golden from the changed source produces a byte-identical file.

### Scenario coverage

Chosen per rendering branch rather than by cross-product — every branch that emits different iPXE gets one scenario:

- every exit type (`sanboot`, `reboot`, `exit`, all three `grub` variants, `refind_efi`, and the unknown-value fallback)
- every arch (`x86_64`, `i386`, `arm64`) and the no-`platform` case
- hidden vs shown menu, and hidden-then-accessed — the only paths through `_chainBoot()`
- registered / pending / unregistered, which select the `pxeRegOnly` set and therefore which rows exist
- both Secure Boot branches (`MOK.der` present and absent)
- both arch-mismatch notices from `_hostOverride()`, and the `_echoSafe()` whitelist
- keymap handling including the Norwegian and Serbian special cases
- simple, nested, and bare webroots

Hook arguments are checked separately: what a plugin receives never appears in the emitted script, so the golden cannot cover it.

### The test has teeth

A golden test that cannot fail is decoration. This one was checked against five deliberate mutations, each of which it catches:

| Mutation | Result |
|---|---|
| menu timeout multiplier `* 1000` → `* 1001` | caught |
| drop `autoboot` from the bootme block | caught |
| sanboot drive order `0x80` → `0x81` | caught |
| weaken the `_echoSafe()` whitelist | caught |
| break the host kernel override | caught |

The two `webroot` hook checks fail against the current `dev-branch` source (`'webroot' was NULL`) and pass against this branch — so that fix is covered, not just claimed.

### Running it

```
php tests/bootmenu-ipxe-output.test.php            # compare against golden
php tests/bootmenu-ipxe-output.test.php --update   # regenerate golden
php tests/bootmenu-ipxe-output.test.php --against <file.php>
```

No database, no network, no PHPUnit — `tests/lib/bootmenu-harness.php` stubs `FOGBase` entirely, matching the existing `tests/route-filter-fields.test.php` convention (plain script, exit 0/1, runs anywhere). Each scenario runs in its own process because several `BootMenu` paths end in `exit()`.

Harness fidelity is where a lazy stub would quietly bake a wrong golden, so three things are deliberate: `getSubObjectIDs('Service', …)` sorts by setting name (which is why `$serviceNames` is written alphabetically — `list()` destructures positionally) and **throws** on an unknown key rather than returning a short array; `fastmerge()` and `arrayInsertAfter()` are copied verbatim from `FOGBase` because their key handling decides menu line ordering; and `resolveHostname()` never does DNS.

## Deliberately not changed

`$initrd` is assigned from `$imagefile` immediately **after** `BOOT_ITEM_NEW_SETTINGS` fires, so a plugin that sets the hook's `initrd` argument has its value discarded three lines later — `$this->_initrd` is built from `$imagefile` regardless. Swapping the two variables at that point is provably a no-op today, which the test confirms.

That is a live bug in the plugin contract rather than dead code, and fixing it changes what third-party plugins can do, so it is left out of a PR whose whole premise is that the emitted bytes do not move. Happy to do it as a follow-up if wanted. The same is true of writes to `webroot`: this PR makes the argument meaningful to read, but nothing downstream reads it back.

## Risk

Low. The four removals are proven byte-identical across 37 scenarios. The `webroot` change can only turn a `NULL` read into a real string — no shipped plugin reads that argument (`location` and `subnetgroup` are the only `BOOT_ITEM_NEW_SETTINGS` consumers, and neither touches it), and writes to it were inert before and remain inert.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DmoFC2e9Cf3fa7mo3HLoN2)_

### DIFF
--- a/packages/web/lib/fog/bootmenu.class.php
+++ b/packages/web/lib/fog/bootmenu.class.php
@@ -308,10 +308,7 @@ class BootMenu extends FOGBase
             . 'chain -ar ${boot-url}/service/ipxe/refind_x64.efi',
             "\n"
         );
-        $reboot = sprintf(
-            'reboot',
-            "\n"
-        );
+        $reboot = 'reboot';
 
         if (isset($_REQUEST['arch']) && stripos($_REQUEST['arch'], 'i386') !== false) {
             //user i386 boot loaders instead
@@ -388,6 +385,16 @@ class BootMenu extends FOGBase
          */
         $bootroot = trim((string)$curroot, '/');
         $curroot = '/' . ($bootroot === '' ? '' : $bootroot . '/');
+        /**
+         * BOOT_ITEM_NEW_SETTINGS passes 'webroot' by reference, but no
+         * $webroot was ever assigned, so PHP created it at the call and
+         * every plugin reading it saw NULL. Bind it to the bare form that
+         * accompanies 'webserver' in the same payload -- the value
+         * 'set fog-webroot' emits -- so the argument means what its name
+         * says. Writes to it are still discarded; nothing downstream
+         * reads the argument back.
+         */
+        $webroot = $bootroot;
         $this->_web = sprintf('%s://%s%s', self::$httpproto, $webserver, $curroot);
         $Send['booturl'] = array(
             '#!ipxe',
@@ -566,7 +573,6 @@ class BootMenu extends FOGBase
                 )
             );
         }
-        $kernel = $bzImage;
         $initrd = $imagefile;
         $this->_timeout = $timeout;
         $this->_hiddenmenu = ($hiddenmenu && !(isset($_REQUEST['menuAccess']) && $_REQUEST['menuAccess']));
@@ -777,7 +783,6 @@ class BootMenu extends FOGBase
      */
     private function _chainBoot($debug = false, $shortCircuit = false)
     {
-        $debug = $debug;
         if (!(isset($this->_hiddenmenu) && $this->_hiddenmenu) || $shortCircuit) {
             $Send['chainnohide'] = array(
                 'set arch ${buildarch}',
@@ -2221,7 +2226,6 @@ class BootMenu extends FOGBase
             $this->_chainBoot(true);
             return;
         }
-        $Menus = self::getClass('PXEMenuOptionsManager')->find('', '', 'id');
         $ipxeGrabs = array(
             'FOG_ADVANCED_MENU_LOGIN',
             'FOG_IPXE_BG_FILE',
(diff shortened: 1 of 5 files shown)


mastacontrola:
Closing: this was opened against the wrong branch first.

Per `CLAUDE.md`, new work belongs on `working-1.6` first and is then ported to `dev-branch` as a separate PR. This went dev-branch-first, which is the ordering that caused trouble on the Secure Boot signing work (#961).

Redoing it on `working-1.6`, where three of these five sites exist:

| Site | `working-1.6` |
|---|---|
| `$kernel = $bzImage` dead store | present |
| `$debug = $debug;` | present |
| `'webroot' => &$webroot` never assigned | present |
| `$reboot = sprintf('reboot', "\n")` | already gone |
| `printDefault()` discarded `find()` | already gone — restructured into `_filterMenus()` + a single `Route::getList` |

The `$initrd`-overwritten-after-`BOOT_ITEM_NEW_SETTINGS` bug noted in the description is present on `working-1.6` too, and will be included there rather than deferred.

A `dev-branch` port will follow once the `working-1.6` change lands.

---

PR #1360, opened 2026-08-25
Make plugins installable over the API

darksidemilk:
Makes plugins installable over the API, and stops the API being able to claim an install that never happened. Those are the same problem from two sides, so they are one PR.

## The problem

Install was reachable only as a management-page POST behind a session and CSRF. The REST API exposed the plugin row as a generic entity, so a client could write `installed` and `schema` directly — the two columns that *record what the installer did*.

`PluginManagementPage::installPost()` shows why that ordering matters: it sets `state`, then calls `Plugin::installdb()` to create the tables, and writes `installed = 1` **last**.

Measured against a real server, on the bundled `location` plugin:

```
PUT /plugin/2/edit  {"installed":"1","state":"1"}   -> 200

GET system/openapi     58 schemas -> 60
                       Location, Locationassociation now present

GET /location          406  SQLSTATE[42S02]: Base table or view not found
GET /location/count    406  SQLSTATE[42S02]: ...
GET /locationassociation
                       406  SQLSTATE[42S02]: ...
```

Setting the column registers the plugin's classes — enough for its routes and schemas to appear in the document — while `schema` stays `0` and no table is ever created. **The server's own OpenAPI document then advertises routes that cannot answer**, and a client generated from it gets commands guaranteed to fail.

It also hides from the obvious repair: `installPost()` filters on `installed IN ('', 0, '0')`, so the **Install** button *skips* a row that already claims to be installed.

## 1. `POST /plugin/{id}/install`

The same three steps as `installPost()`, in the same order — activate, `installdb()`, then `installed = 1` last. Activation blockers are refused first, the same gate the page applies, so the API is not a way around a plugin the server has a reason to refuse.

`installdb()` is called **unconditionally** rather than only when the plugin is not yet installed. Migration steps are append-only and idempotent by contract (`docs/PLUGIN_SCHEMA_MIGRATIONS.md`) and `Schema::applyUpdates()` resumes from the stored count, so calling it on an installed plugin applies only steps it has not seen — which is what the UI calls **Upgrade**. One route therefore installs, *and* brings a plugin whose code ships newer `schema()` steps up to date, without a drop and recreate.

That also makes it the repair for a row broken the way the one above was — which the Install button structurally cannot fix.

Permission is **`plugin.install`**, the node the web Install button checks, rather than `plugin.edit`: running a plugin's schema migrations is a different authority from editing its row. Declared in `API_ROUTE_PERMISSIONS` rather than relying on the fail-closed default in `resolveApiPermission()`.

## 2. `plugins.installed` and `plugins.schema` become server-owned

One entry in `Route::$serverOwnedFields`. That list already drives `_refuseServerOwned()`, which answers **400 only when the value would change**, so reading a plugin and sending the object back unmodified still works. `openapi.class.php` reads the same list, so the document marks both `readOnly` with the standard server-owned description.

**`state` is deliberately not included.** Activating is a column write and nothing else — `installPost()` and the Activate action both just set `state = 1` — so a client writing it does the whole job rather than half of it. Activation over the API keeps working exactly as before.

## Verification

Regenerated the document with `spec/tools/dump-openapi.php`:

```
/plugin/{id}/install   operationId=pluginInstall
                       responses 204/400/401/403/404/500
                       x-fog-permission=plugin.install

Plugin.installed  readOnly=True
Plugin.schema     readOnly=True
Plugin.state      readOnly=None
```

373 paths, up from 372. `php -l` clean on all three files; added lines within the 80-column style.

**Not yet exercised at runtime.** I have not run `POST /plugin/{id}/install` against a server built from this branch — the reproduction above was done against a stock server before the fix. The refusal half rides on the pre-existing `_refuseServerOwned()` path, unchanged here and already exercised by the `task`/`host`/`user` entries; the install handler is new code and deserves a real run. Happy to do that once it is deployed somewhere, or tell me if you would rather it come with a test.

### DIFF
--- a/packages/web/lib/fog/authorization.class.php
+++ b/packages/web/lib/fog/authorization.class.php
@@ -222,6 +222,10 @@ class Authorization extends FOGBase
         'pendingmacs' => 'host.view',
         'snapinCreateWithFile' => 'snapin.create',
         'uploadSnapinFiles' => 'snapin.create',
+        // The node the web Install button checks, rather than plugin.edit:
+        // installing runs a plugin's schema migrations, which is a
+        // different authority from editing its row.
+        'pluginInstall' => 'plugin.install',
         'settingsCacheView' => 'settings.view',
         'settingsCacheFlush' => 'settings.edit',
         'settingsCacheRefresh' => 'settings.edit'
--- a/packages/web/lib/fog/openapi.class.php
+++ b/packages/web/lib/fog/openapi.class.php
@@ -2009,8 +2009,73 @@ class OpenAPI extends FOGBase
             '/storagegroup/{id}/uploadsnapinfiles' => [
                 'parameters' => [self::_idParameter()],
                 'post' => self::_uploadSnapinFilesOp()
+            ],
+            '/plugin/{id}/install' => [
+                'parameters' => [self::_idParameter()],
+                'post' => self::_pluginInstallOp()
             ]
         ];
+    }
+
+    /**
+     * POST /plugin/{id}/install.
+     *
+     * Written out here rather than falling out of the generic shapes for
+     * the same reason the upload routes are: it is an action, not CRUD on
+     * a row. The generic edit route cannot express it, and deliberately
+     * refuses to pretend it can -- plugins.installed and plugins.schema
+     * are server-owned, because both record what this operation DID.
+     *
+     * @return array
+     */
+    private static function _pluginInstallOp()
+    {
+        return self::_op(
+            '',
+            'pluginInstall',
+            _('Install a plugin'),
+            _('Activates the plugin, applies its schema migrations, and '
+                . 'records the result -- the same three steps, in the same '
+                . 'order, as the Install action in the web UI. Idempotent: '
+                . 'migration steps are append-only and are resumed from the '
+                . 'count already applied, so calling this on an installed '
+                . 'plugin applies only steps it has not seen, which is what '
+                . 'the UI calls Upgrade. Setting `installed` through the '
+                . 'generic edit route instead is refused: that column '
+                . 'records that this operation succeeded, and asserting it '
+                . 'without running the migrations leaves the plugin\'s '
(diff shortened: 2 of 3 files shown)


mastacontrola:
Superseded by #1363, now merged to `working-1.6`. Your two commits went in unchanged and are attributed to you — both halves of the problem were identified correctly, and the reasoning in the commit messages is what made the rest quick to find.

I ran it against a server before merging, which was the outstanding bit. Four things came out of that:

**1. The blocker gate was defeated by an unrelated key in the request body.** `Plugin::activationBlockers()` lists plugins through `Route::getList()`, and `listem()`'s `$inputoverride` blanks only one of the two places it reads `php://input`. The other is `getsearchbody()`, which turns any of the class's own fields found in the body into a `WHERE`. On a plugin declaring `fog_min 9.9.9`:

```
POST /plugin/2/install                   -> 400 needs FOG 9.9.9
POST /plugin/2/install {"name":"ldap"}   -> 204, installed
```

Fixed in `listem()` rather than in `activationBlockers()`, because any gate built on a `getList()` was defeatable the same way.

**2. `installdb()` is idempotent only for a manager that adopted `schema()`.** Without one it falls back to the legacy `install()`, which opens with `uninstall()` — a DROP. The Install button's `installed IN ('', 0, '0')` filter was what had been holding that off, so dropping the filter deliberately meant "install an installed plugin" silently meant "empty it". A second install of such a plugin now answers 400; a first still works. `taskstateedit` and `tasktypeedit` still ship no `schema()`.

**3. The error arms sent a bare string in a body labelled `application/json`** while the document declared the `Error` schema. Now `setErrorMessage()`.

**4. Leaving `state` writable was right, but it wanted the same blocker gate** — an installed plugin deactivated because an upgrade moved past its `fog_max` could be switched back on with one PUT. `edit()` now applies it, on the transition to active only.

Also tagged the operation `plugin` rather than `system`, the way the two upload ops do. Full detail in #1363.

PR #1383, opened 2026-08-26
Use MAC of first link-up interface instead of forcing net0 for setmacto

MrLego8-9:
## What
The iPXE scripts always set the `setmacto` kernel parameter to the MAC
address of `net0`, regardless of which interface actually has a link.
This patch changes that so `setmacto` uses the MAC address of the first
NIC that has an active link during iPXE's network autoconfiguration,
falling back to the same behavior as before (`net0`) when only one NIC
is present or when `net0` itself is the one with a valid link.

## Why
On multi-NIC machines where the interface labeled `net0` by iPXE has no
link (unplugged) and a different interface (e.g. `net1`) does, iPXE still
autoconfigures and gets a DHCP lease over the interface that has a link.
However, the kernel is told via `setmacto` to apply `net0`'s MAC address
to all interfaces. If only the actually-connected NIC's MAC is
provisioned for DHCP on the network, FOS then fails to get an IP,
because every interface presents the MAC of the unplugged one.

## Reproduction steps
1. Take a machine with 2+ NICs.
2. Leave the interface iPXE enumerates as `net0` unplugged, and plug in
   another (e.g. `net1`).
3. Configure DHCP/network access to only recognize the MAC of the
   plugged-in interface.
4. PXE boot into FOS.
5. Expected: FOS gets an IP over the interface with link.
   Actual: FOS fails to obtain an IP, because `setmacto` forces every
   interface (including the one with link) to use `net0`'s MAC.

## Testing
Tested on 60+ physical machines:
- Multi-NIC machines with `net0` unplugged: FOS now gets an IP over the
  linked interface.
- Single-NIC machines: behavior unchanged.

## Environment
- FOG version: 1.5.10.2254

### DIFF
--- a/packages/web/lib/fog/bootmenu.class.php
+++ b/packages/web/lib/fog/bootmenu.class.php
@@ -395,6 +395,16 @@ class BootMenu extends FOGBase
          */
         $webroot = $bootroot;
         $this->_web = sprintf('%s://%s%s', self::$httpproto, $webserver, $curroot);
+        /**
+         * setmacto is the MAC FOS forces onto whichever interface it manages
+         * to reach us on, so it has to be the MAC iPXE actually booted with.
+         * ${net0/mac} was wrong on any machine whose first enumerated NIC has
+         * no link: iPXE gets its lease over the NIC that does, FOS then
+         * rewrites that NIC to the unplugged one's MAC and the re-DHCP fails.
+         * ${netX} is iPXE's alias for the last opened network device, so it
+         * follows the interface that got us here and still resolves to net0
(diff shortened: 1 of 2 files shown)


mastacontrola:
Thanks for this, the bug is real and the repro is spot on. I pushed a small change to your branch: `${${ifname}/mac}` does give the right answer, but only because an unscoped iPXE lookup resolves through the netX redirect block anyway, so `${netX/mac}` is the same value in one expansion and is iPXE's documented name for the device it booted from. Same end goal, just the more proper spelling. I also ported it to working-1.6 in #1384.

PR #1393, opened 2026-08-26
Allow elliptic curve root CAs

osiktech:
this patch allows the usage of root CAs using elliptic curves over RSA. 

I have at least one installation with a root CA using elliptic curves which failed with the current installer.

### DIFF
--- a/docs/EXTERNAL_CA_AND_LETSENCRYPT.md
+++ b/docs/EXTERNAL_CA_AND_LETSENCRYPT.md
@@ -166,6 +166,20 @@ The installer (working-1.6 and later) can sign FOG's server certificate with a C
 | `--ca-key`  | Intermediate CA **private key** (PEM) | Must match `--ca-cert` |
 | `--ca-root` | **Root** CA certificate (PEM) | `--ca-cert` must verify against this |
 
+**Key algorithms.** Any key OpenSSL can load is accepted -- RSA, RSA-PSS, EC on
+any curve, Ed25519, Ed448 -- because the installer pairs `--ca-key` to
+`--ca-cert` by comparing public keys rather than RSA moduli. The one refusal is
+**DSA**, which TLS 1.3 removed and no current browser trusts.
+
+Accepted is not the same as usable everywhere, though, and the narrow verifier
+is iPXE's. The imported CA is compiled into the iPXE binary, and iPXE verifies
+**RSA and ECDSA P-256/P-384 signatures only** -- no P-521, no Ed25519/Ed448, no
+RSA-PSS. A CA outside that set serves the web UI and satisfies fog-client
+normally, then fails HTTPS netboot at `boot.php`; the installer prints a note
+saying so at import time. See [How iPXE validates
+HTTPS](#how-ipxe-validates-https). **If you netboot over HTTPS, use RSA or EC
+P-256/P-384.**
+
 Enable it with `--external-ca`, or answer the interactive prompt during install:
 
 ```bash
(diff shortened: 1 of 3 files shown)


osiktech:
@mastacontrola could you please check the PR? It should not break anything as it just introduces a way to allow different  encryption algorithms used in the certificates. What do you think?

darksidemilk:
I can take a gander.

darksidemilk:
Pushing some changes to your fork shortly to expand upon this a bit

darksidemilk:
Thanks for this — the core observation is right and worth having: for an EC certificate `openssl x509 -noout -pubkey` and `openssl ec -pubout` really do produce byte-identical PEM, so the pairing works. I've pushed three commits onto your branch rather than asking you to round-trip them; here's what changed and why.

**The allow-list had to come out.** `openssl x509 -text` prints `Public Key Algorithm: rssaPss` for an RSA-PSS CA, which matches neither arm of the `case`, so the `*)` arm's `exit 1` aborted the install. That's a regression: before this patch, `openssl rsa -noout -modulus` reads an RSA-PSS key fine, so those CAs paired and installed. Ed25519 and Ed448 hit the same wall — and those are elliptic-curve CAs, which is the thing the PR sets out to support.

**What replaced it.** One comparison, no algorithm detection at all:

```bash
certpub=$(openssl x509 -pubkey -noout -in "$certsrc" 2>>$error_log)
keypub=$(openssl pkey -pubout -in "$keysrc" 2>>$error_log)
```

`openssl pkey -pubout` emits byte-identical SPKI PEM to `openssl x509 -pubkey` for every algorithm OpenSSL can load. I checked against RSA, EC P-256/P-384/P-521, an `ecparam -genkey` key with its leading `EC PARAMETERS` block, RSA-PSS, Ed25519, Ed448 and DSA — identical in all eight. Both of FOG's leaf-signing shapes (`_createWebLeaf`, which passes no digest flag, and `fog-sign-node-cert`, which passes `-sha256`) sign correctly under all eight too. So there's nothing for a list to earn here; it can only fall behind OpenSSL again the way it just did with RSA-PSS.

**DSA stays refused, deliberately.** It's the one thing OpenSSL will pair that nothing downstream accepts — TLS 1.3 removed DSA signatures outright, no current browser trusts a DSA chain, and iPXE has no DSA at all. It was already being rejected before your patch, just dishonestly: `openssl rsa -modulus` can't read a DSA key, so the installer said "the supplied CA private key does not match the supplied CA certificate", which isn't true. Now it says DSA and says why.

**One thing worth knowing about your EC CA.** iPXE's verifier is narrower than OpenSSL's, and it's this certificate that it gets — `_resolveIpxeTrust()` hands `${PKI_web_ca_cert}` to `buildipxe.sh` as `CERT=`/`TRUST=`. The pinned tag carries `crypto/rsa.c` plus `ecdsa.c` with `p256.c` and `p384.c`, and nothing else: no P-521, no EdDSA, no RSA-PSS. A CA outside that set serves the web UI fine and satisfies fog-client, and then HTTPS netboot dies at `boot.php` with iPXE's "Permission denied" out of `x509.c`, with nothing server-side pointing at the cause. The installer now prints a note at import time instead of leaving that to be discovered at a PXE client. It's advisory, not a rejection — a site that netboots over HTTP is unaffected, and the limit is a pinned iPXE version rather than anything about the CA. **If you netboot over HTTPS, keep the CA on RSA or EC P-256/P-384.**

**Smaller things folded in:** `$certsrc` is quoted and stderr goes to `$error_log` on the new openssl calls, matching every other call in that function — unquoted, a CA path containing a space made the algorithm come back empty and aborted an otherwise valid RSA install, and it got past the readability guard above because *that* one is quoted. The new locals are declared. The failure branch now `echo "Failed"`s first so it closes the `dots` progress line the way its three siblings do.

Plus a regression test (`tests/external-ca-key-algorithms.test.sh`) that generates a root and intermediate on each of the nine algorithms and runs `validateExternalCA()` over them. Against the code as it stood before these commits, the RSA-PSS, Ed25519 and Ed448 cases fail with "does not use a supported algorithm" — so it's actually holding the regression shut rather than passing either way. It also pins that DSA stays refused *and names DSA*, and that an EC certificate paired with a different EC key on the same curve still fails, so the comparison can't quietly become always-true.

And `docs/EXTERNAL_CA_AND_LETSENCRYPT.md` no longer tells admins to debug a pairing failure with `openssl rsa -noout -modulus`, which was wrong for exactly the CAs this PR adds.

Not verified end to end: no install was run against a real EC CA. The pairing, signing and chain-verify steps are exercised on fixtures; Apache/nginx actually serving an EC-signed leaf, and iPXE actually verifying one, are not.

PR #1409, opened 2026-08-27
Answer with a list, not a bare array

darksidemilk:
**This is the blocker.** 108 operations answer with a bare top-level JSON array, and no code generator can model that — so no generated client of FOG's API compiles, in any language. Unlike #1408 (which is cosmetic), nothing downstream works until this lands.

Until now nobody had tried. Generating FOG's document has always reported success — but *generating* and *compiling* are different gates, and only the first was ever checked. The C# does not compile: **333 errors, from a run with zero warnings.**

AutoRest builds no model for a top-level array, falls back to the only named schema on the operation (`Error`), then writes `.Count`, `[0]` and `foreach` against it. Its own comment on the generated line:

```csharp
// (await response) // should be FogApi.Models.IError
// response should be returning an array of some kind.
```

## Measured before changing a wire format

Naming the schema was tried first, and does not work. Five response shapes were generated and compiled:

| shape | result |
|---|---|
| inline array, `items: {type: object}` | ✗ |
| `$ref` to a **named** top-level array schema | ✗ |
| inline array whose items are a named object | ✗ |
| inline array of integers | ✗ |
| **an object wrapping the array** | ✓ compiles |

So there is no document-only fix. The rows have to sit under a property — which is where every list route already puts them. `/list` has always returned an envelope with `data`; these routes were the inconsistent ones.

## What changed

- `/{class}/ids` and `/{class}/names` — 104 routes, now `{"data": [...]}`
- `/availablekernels`, `/availableinitrds` — now `{"data": [...]}`

Two more were not wire changes at all, but the document lying about its own server:

- **`/pendingmacs`** delegates to `Route::listem()`, so it has always answered with a full page. It was documented as a bare array. Now points at `MACAddressAssociationPage`.
- **`POST /{class}/join`** was documented as returning an array of ids. It never has — `Route::joining()` ends at `sendResponse($code)` with no body, so a generated client was told to wait for a list that never arrives. Now documented as `201` with no content.

`/system/export` also loses `format: binary`. AutoRest cannot model a binary response and emits `File.WriteAllBytes(path, result)` with `result` typed as the error schema. The content type is irrelevant — `application/sql`, `application/octet-stream` and `text/plain` all compile once `binary` is dropped, and a SQL dump is text, so a plain string is the honest declaration.

## Consumers

One first-party caller reads a changed route. `fog.host.list.js` drives the group typeahead on the Host list page from `/group/names`, doing `$.map(data, …)` and `data.length`. Updated — and written to accept **both** shapes, so the page keeps working against a server that has not been updated yet.

## Result

Generated, compiled and imported, which had never been done before:

| | before | after |
|---|---|---|
| compile errors | 333 | **0** |
| `FogApi.private.dll` | — | builds, 20.6 MB |
| module import | — | **474 commands**, all `Verb-Fog*`, no collisions |
| `Get-FogHostId` / `Get-FogHostName` | uncompilable | exist |
| `Get-Help Get-FogHost` | — | returns FOG's own descriptions |

Two `multipart/form-data` routes remain ungeneratable and are excluded client-side. That one is **not** fixable here — a minimal probe (one object, one `file` property) fails identically, so AutoRest cannot do multipart at all.

Independent of #1408; both touch `openapi.class.php` and may need a trivial rebase depending on merge order.


darksidemilk:
### Open question for review: the `/system/export` half of this PR

Flagging a tradeoff in this PR that I do **not** think should be settled by me. Everything else here is a straightforward fix; this one route is a judgement call about FOG's document versus a generator bug.

**What this PR does to it:** drops `format: binary` from the `200` on `GET /system/export`.

**Why:** with `format: binary`, AutoRest emits its file-download path — a mandatory `-OutFile` and `File.WriteAllBytes(path, result)` — but types `result` as the *error* schema, so it does not compile. That happens whenever a binary `200` sits beside an error response, which every FOG route has. Measured across five variants: content type is irrelevant (`application/sql`, `application/octet-stream`, `text/plain` all behave the same); the presence of `format: binary` is the trigger.

**What it costs:** the generated cmdlet changes shape.

| | `-OutFile` | writes | compiles |
|---|---|---|---|
| today (`format: binary`) | mandatory | file, streamed to disk | **no** |
| with this PR | **gone** | whole dump to the pipeline as one string | yes |

For a database dump that is a real downgrade: the entire dump is materialised in memory, there is an encoding risk on a payload that must round-trip byte-exact, and an `Export-` verb that writes no file is misleading.

**The alternative** is to keep `format: binary` here — it is the truth, this route streams a file — and instead exclude the operation in the client's own generator config, which is how the two `multipart/form-data` routes are already handled (AutoRest cannot do multipart at all). FOG's document then stays accurate for every language client, and the one route AutoRest cannot express is reached through a raw-request cmdlet.

I have left the change in place so the rest of the PR can be evaluated and the module keeps building, but I would rather the FOG side decide whether weakening the declaration is acceptable. Happy to split `/system/export` out into its own change either way.

The other 107 operations in this PR are unaffected by this question.

PR #1451, opened 2026-08-29
Record what each host reports about Secure Boot, and what was enrolled

mastacontrola:
## The problem

ADR 0008 scopes the Secure Boot enrolment task to machines that are **not currently enforcing Secure Boot**, and states why in the task's own description: an enforcing machine does not trust this server's kernel, so it cannot boot FOS, so it cannot run the task that would make it trust us.

That constraint was written down and then had nowhere to live. Nothing in FOG recorded what state any machine was in, so finding the eligible ones meant an administrator guessing — and a wrong guess is not a clean failure. It schedules a task against a host that PXE boots, refuses the kernel and reboots. What the administrator sees is a host that did not image, which looks like a broken image or a broken network, not like a task that was never eligible.

FOS already computes the answer in `sbState()`, but FOS only runs when someone schedules a task, and the whole problem is deciding what to schedule. A host stays unknown until FOS runs, which may be months away or never — the same shape as architecture before schema step 369.

iPXE runs on **every** PXE boot and can already read it.

## What this does

`default.ipxe` now posts `${efi/SecureBoot}` and `${efi/SetupMode}` alongside the platform, arch, product, manufacturer, sysuuid and MACs it already sent. That file is generated by the installer, so this is two lines in `configureDefaultiPXEfile()` — **no fog-ipxe change and no FOS change** for the observed half.

Two halves, kept apart because they are different kinds of fact:

| | Observed (`hostSbState`, `hostSbStateTime`) | Asserted (`hostSbEnrolled`, `hostSbEnrollCert`, `hostSbEnrollVia`) |
|---|---|---|
| What it is | what a machine said about itself | a record that somebody enrolled something |
| Written by | iPXE on every PXE boot; FOS when the task runs | FOS for `db`/`trusted`/MOK; a human for the USB path |
| Editable | **never**, in the UI or over the API | **yes**, and it has to be |
| Timestamp | stamped by the server from its own clock | the date of the enrolment, which may be historical |

The observed half is never editable because there is no correct manual value for an observation. Enforcement is `Route::$serverOwnedFields['host']` (400 on write, pinned by `tests/api-server-owned-fields.test.php`) plus **disabled**, not readonly, inputs — a disabled input is not submitted at all.

## HARD: neither half is a security control

`service/ipxe/boot.php` is unauthenticated by necessity — a booting NIC has no credential — so everything in the observed half is attacker-controlled, and the asserted half is typed by a human. **Both exist for targeting, filtering and display. Nothing may ever read either as a security control.** This is stated as a hard constraint in ADR 0029.

What makes it safe is the asymmetry of what a lie buys:

- Spoofing `disabled` (or leaving the state unknown) gets a host a task that cannot work. That is a wasted task — exactly what happens today, on every host, with no record at all. Nothing regresses.
- Spoofing `enforcing` gets a host *excluded* from an enrolment it is denying itself. It costs nothing.

The boot decision keeps reading the live request and never this column, exactly as step 369 already requires for `hostArch`; `_filterMenus()` keeps gating on the live `${platform}`.

## Six states, not a boolean

Five are taken verbatim from FOS's `sbState()` so the two reporters cannot drift into two vocabularies for one fact. `unknown` is the sixth and is server-side only.

| | |
|---|---|
| `unknown` | nothing has reported yet |
| `nonefi` | booted BIOS/CSM |
| `noefivars` | UEFI, variables unreadable |
| `setup` | Setup Mode — `db` is writable, enrolment is unattended |
| `enforcing` | Secure Boot ON — the task cannot run |
| `disabled` | Secure Boot OFF — the ADR 0008 case |

**`setup` and `disabled` are not collapsed into one "off".** Turning Secure Boot off leaves the platform key in place, so `db` still refuses a write; only Setup Mode accepts one. `fog.enrollsb` branches on exactly that, and it is the difference between an enrolment that finishes with nobody at the keyboard and one that needs a human at MokManager — the single most useful thing to know before scheduling.

`unknown` is likewise not a synonym for anything. Collapsing it into `disabled` is the dangerous direction, because `disabled` is what makes a host look like a valid target.

## The format was measured, not read out of the docs

`SecureBootState::fromBootRequest()` is the only code that understands the wire format, and `tests/secureboot-state-classifier.test.php` is the only place it is asserted. Measured on QEMU/OVMF:

| Rig | `platform` | `${efi/SecureBoot}` | `${efi/SetupMode}` |
|---|---|---|---|
| `OVMF_CODE.secboot.fd` + plain `OVMF_VARS.fd` | `efi` | `00` | `01` |
| firmware with no Secure Boot support | `efi` | *(empty)* | *(empty)* |
| SeaBIOS + FOG's own `ipxe.lkrn`, real `default.ipxe` | `pcbios` | *(empty)* | *(empty)* |

Two consequences that are load-bearing and not obvious:

**The legacy BIOS path cannot break.** `EFI_SETTINGS` is gated on `PLATFORM_efi`, so the `efi` scope is never registered on a BIOS build; iPXE substitutes an empty string for a setting it cannot resolve rather than erroring; and `param` accepts a key with no value. Measured, not reasoned — FOG's own `ipxe.lkrn` ran this exact `default.ipxe` through to the chain. The new params are also emitted **above** the `net1..net7` chain that short-circuits to `:bootme`, and `tests/netboot-host.test.sh` asserts that ordering.

**An absent param and an empty param are different facts.** The captured BIOS body was `…&platform=pcbios&secureboot=&setupmode=` — present and empty, so `filter_input()` returns `''`. A server that has not re-run the installer sends neither and returns `null`. That is what lets a not-yet-upgraded server read as `unknown` instead of masquerading as legacy BIOS.

## The payoff

The enrolment task now refuses a target it cannot run on, turning ADR 0008's constraint from something a technician remembers into something the server checks.

**Written at two call sites because there are two code paths.** A non-imaging group task never calls `Host::createImagePackage()` — it batch-inserts over the member ids — so `Group::createImagePackage()` carries its own check. It **filters** rather than throwing: a group is a mixed bag by nature, ADR 0008 gives the task `ttIsAccess='both'` precisely so a batch can be scheduled at once, and refusing a whole group because one member came back with Secure Boot already on would make the group path useless for the case it exists for. Both call the same `SecureBootState::refusalReason()`.

**`unknown` is allowed through, deliberately.** Nothing is reported until a host PXE boots, so refusing it would make the task unusable on every existing fleet until a full boot cycle had happened — in exchange for preventing the one thing that already happens today, a wasted task. The host list column and its badge are what keep an unknown host from *looking* eligible.

## Also in here

- **A staged MOK is not an enrolment.** `fog.enrollsb` has three exits and all three end with the same argument-free completion POST, so the server cannot tell them apart. `service/secureboot.report.php` takes the explicit report. Recording "enrolled" for a staged MOK is a lie an administrator acts on: they turn Secure Boot on and the machine stops booting. `mok-pending` is its own value and the grid renders it as pending rather than as a date. `trusted` is likewise its own value and not a synonym for `db` — it means the machine already trusted the certificate and nothing observed how it got there.
- **The record references what was enrolled.** `hostSbEnrollCert` holds the certificate's SHA-256. An enrolment date alone goes stale in silence — FOG has PKI zones, a multi-server CA and certificates that expire, so "enrolled 2026-03-14" says nothing about whether the machine trusts what this server serves today. Costs one column and no new computation: `FOGConfigurationPage::secureBoot()` and FOS's `sbCertFingerprint()` already produce the identical string, so the check is equality.
- **Last write wins, provenance is shown rather than arbitrated.** A technician correcting a wrong machine-written record is legitimate, and a rule letting the machine outrank the human would create a value nobody could fix. `hostSbEnrollVia` is rendered beside the date instead.
- **`hostLastPing` and `hostLastCheckin` added to `Route::$serverOwnedFields`.** The host form has always disabled those inputs and its comment has always called them server-owned, but the API never refused them. A rule enforced by one form is a rule that holds until someone adds a second writer.
- `generateIpxeItems()` excludes all five columns. FOG's record of what a machine reported has no business being echoed back at that machine.

## Size and reversibility

| | |
|---|---|
| Schema | steps **376** and **377**, five additive NULL-able columns on `hosts`, no data migration. Separate steps so a partial failure leaves an unambiguous state (same reasoning as 369/370). `FOG_SCHEMA` 375→377, `FOG_BCACHE_VER` 321→322 |
| Installer | two lines in `configureDefaultiPXEfile()` |
| New files | `src/Boot/SecureBootState.php`, `service/secureboot.report.php`, two tests, one ADR |
| Reversible | yes — the columns are additive and NULL-able, and a server that never re-runs the installer simply reads `unknown` |
| Lab run | QEMU/OVMF for the wire format (done). An end-to-end `enforcing` reading needs the signed chain on the VirtualBox rig; the classifier is built so being wrong about it fails **closed** |

## Companion change

`FOGProject/fos` needs `fog.enrollsb` to report which of its three exits it took. Until that ships, the observed half works on its own and the enrolment record is hand-entered only.

## Downstream

No change to `Route::$validClasses`, so FogApi's hardcoded class list is unaffected.

## Verification

- `tests/run-all.sh` — **185 passed, 0 failed**
- `vendor/bin/phpstan analyse` — no errors
- `vendor/bin/phpstan analyse -c phpstan-tests.neon` — no errors
- Both new gates mutation-tested: every assertion was made to go red by reintroducing the defect it exists to catch. That found a real gap on the first cut — `refusalReason()`, which both refusal sites call, had no coverage at all, and a mutation making it return `''` for an enforcing host passed green.

https://claude.ai/code/session_01C2gFRQ3Khhx24FNYMZhW6N


mastacontrola:
## Added: the comparison the fingerprint column exists for

Half 2 was storing `hostSbEnrollCert` and never comparing it. ADR 0029 decision 5 says the column is there because "was this machine ever enrolled" is not the question an administrator has — it's "does this machine trust the certificate I am serving today" — and it ends "the check is string equality." Nothing performed it. Stored, rendered, exported; the question left to somebody checking 95 hex characters against a different page by eye.

`SecureBootState` now owns the certificate half alongside the state half:

| | |
|---|---|
| `normalizeFingerprint()` | one format, one place. It was written out longhand in **four** files |
| `serverFingerprint()` | memoised, so the grid hashes the DER once per request, not once per row |
| `enrolmentFreshness()` | `current`, `stale`, or `''` |

`''` — not `stale` — for all three cases where the question cannot be answered: nothing recorded, an unparseable stored value, or a server with no signing certificate. None is evidence a machine trusts the wrong key, and reporting them as stale would put a red badge on every host in a fleet that has simply never enrolled. A warning that is on everything is a warning nobody reads.

**Surfaced in the two places the answer is wanted.** The host form grows a read-only Certificate Status row, rendered only when it has something to say, computed from the *stored* value rather than `INPUT_POST` — a rejected post re-renders what was typed, and comparing that would report the freshness of a value the database does not hold. The host grid gets a derived `sbenrollfresh` column, the same headerless shape `sbstatecode` uses, badging only the stale rows. The badge is client-side per GH-1446: this column is in the host export and every escaping consumer would print a `<span>` as literal text.

`FOGConfigurationPage::secureBoot()` now reads its SHA-256 through the same helper rather than computing it inline, so the string on that page and the string each host is compared against are the same string *by construction*. They agreed when both were written longhand; "they agree today" is not the same property as "they cannot disagree".

## Also recorded: why `mok-pending` gets no automatic resolution

Nothing observes the human at the MokManager screen, so a pending record does not clear itself — and it deliberately gets no inference to make it. FOS's `sbCertTrusted()` already answers the question definitively, checking **both** `db` and MokList, so re-running the enrolment task on a confirmed host reports `trusted` and the record corrects itself. That is an observation. Anything else here would be a guess, and guessing is what decision 6 exists to stop.

## Verification

`FOG_BCACHE_VER` 322 → 323 for the changed host list JS. Suite 185/0, both phpstan passes clean, `route-column-contract.txt` regenerated for the one new derived column.

Eight mutations run against the new assertions, eight caught — including a normaliser that passes its input through unchanged, and a freshness that reports every host as current.

PR #1474, opened 2026-08-29
Card header descriptions render below the title, not welded to it

mastacontrola:
A card header that carries a description renders it welded to the title, with no separator at all:

> **Import Reports**This section allows you to upload user defined reports that may not be a part of the base FOG install.

AdminLTE 4 sets `.card-title { float: left }` so a header can carry `.card-tools` beside it. Twelve headers also emit a `<p class="form-text">` description in the same header, and a block element following a float wraps **around** it — so the paragraph starts on the title's own line.

Two of the twelve come from shared helpers (`FOGPage::renderAssocCreate`, `FOGPageRender`), so it reads that way on every page they serve. Found while screenshotting Import Reports for the [reports docs rewrite](https://github.com/FOGProject/fog-docs/pull/151).

## The fix

```scss
.card-header > .form-text {
  clear: both;
  display: block;
  margin-top: 0.25rem;
}
```

**The description clears; the title keeps its float.** Unfloating `.card-title` would fix these twelve headers by moving the `.card-tools` button rows on *every* card in FOG — a far bigger change than the one being made. Direct child, so it reaches a header's own description only, and it sits in the existing `.app-main` block, so the login page's separate asset list is untouched.

`FOG_BCACHE_VER` 329 → 331. Hygiene rather than correctness: without it a browser serves the stale stylesheet until a hard refresh or until the entry expires — an annoyance for whoever hits it, not a broken install. Worth one digit so nobody has to know to press ctrl-shift-R. 331 rather than 330 because #1471 landed a different edit to the same stylesheet under 330 while this branch was open.

## Verification

Live, in a browser against a 1.6 web root:

| Check | Result |
|---|---|
| Import Reports description | now below the title (`p.top` 132 = `title.bottom` 132) |
| Client Management (2 description headers) | both below the title |
| `.card-tools` headers on FOG Configuration, Client Management, Dashboard | all 4 still inline with their titles |

**`tests/card-header-description.test.php` — 6 checks — pins the pair, not the rule.** Nothing in the build compiles `fog-default-ui.scss` into the `.min.css` a browser actually loads, so editing the source and forgetting to regenerate passes every other check in the tree and reaches nobody. That is the silent failure the gate exists for.

**Five mutations, all red:** dropping the rule from the SCSS; `clear: both` → `clear: none`; leaving the minified file stale; reverting `FOG_BCACHE_VER`; and "simplifying" it by unfloating `.card-title`.

Regenerated with `sass --style=compressed`. The new minified artifact differs from its predecessor by **exactly the one added rule, byte for byte** — checked before committing, so no toolchain skew rode along with it.

`tests/run-all.sh`: 201 passed, 0 failed. Both phpstan passes clean.

## Downstream

None — no route or class change.


mastacontrola:
Updated for `working-1.6` after #1471 landed.

Two things that needed handling rather than a plain merge:

- **`FOG_BCACHE_VER` is now 331, not 330.** #1471 landed a different edit to `fog-default-ui.min.css` under 330 while this branch was open, so 330 no longer distinguishes that stylesheet from this one — a browser holding #1471's copy at `ver=330` would serve it back rather than fetch the one carrying this fix. The gate moved with it, and reverting to 330 makes it red.
- **The minified stylesheet was re-verified, not trusted.** git text-merged a one-line file, which is exactly the shape that can silently drop content. Recompiled from the merged SCSS and compared: the merged artifact was already byte-identical to a fresh `sass --style=compressed` run, so nothing was lost and nothing needed re-committing.

`tests/run-all.sh`: 202 passed, 0 failed. Both phpstan passes clean.

PR #1497, opened 2026-08-30
Stop Responsive re-measuring every loaded row on every column adjust

mastacontrola:
## Problem

The host list still took ~1.5s to settle for 86 rows after #1486, and got worse with list size. The server was never involved — the grid POST answers in **30ms** by curl.

The cost is `DataTables.Responsive._resizeAuto()`. It answers **every** `column-sizing` event — which is every `columns.adjust()` — by deep-cloning the header, the footer and **every row of the current page** into a hidden table so the browser will compute each column's natural width. With Scroller the "current page" is every row fetched so far, so the cost is linear in the size of the list.

Measured on 10.255.20.1, host list, 86 rows:

| | |
|---|---|
| `_resizeAuto` calls per load | **8** |
| total | **~870ms** |
| per call | ~140ms |
| of which building the clone | **15ms** — the other ~125ms is forced layout |

It scales at roughly **10ms per row per page load**: 2 rows 23ms, 4 rows 39ms, 29 rows 194ms, 86 rows 870ms. A 1000-host list implies ~10s.

A note for anyone re-measuring this: the same grid POST that curl answers in 30ms *appears* as 1062ms to a hooked `XMLHttpRequest`, purely because `loadend` cannot fire while the main thread is blocked. Prove server time with curl, not from the page.

## Fix

**Five of the eight calls returned a `minWidth` vector byte-identical to the call before them** — 551ms provably recomputing an unchanged answer.

The widths come only from the markup Responsive clones, and it measures with the table at `width:auto` inside a 1px hidden box, so the viewport is not an input. So sign exactly what gets cloned — the table node's class list, the visible column set, and the header/body/footer HTML — and restore the previous answer when the signature is unchanged. The signature costs **~1.5ms** against the ~140ms it skips.

`fogMemoizeResponsiveMeasure()` in `fog.common.js`, installed once from `$.fn.registerTable()`.

### Why the cache key is what it is

- **Not the draw counter or row count.** Responsive converges over *two* passes within a single draw: the first measures, `_resize()` then adds `dtr-control` to the first column, and the second legitimately measures it 17px wider (166 → 183). A key coarse enough to treat those as one call would cache the pre-convergence width and leave that column permanently narrow.
- **The class list is sorted before signing.** `addClass`/`removeClass` reorder the tokens without changing what renders, and that alone defeated the cache twice per load.
- **Dropped on `window.resize`.** The markup does not change when the browser zooms or the font size does, but the metrics do.
- `fog-table-fixed` stays in the signature — it is `table-layout: fixed`, which genuinely changes how the clone lays out.

## Verification

The real patched file was served over CDP against the live server (`Fetch.fulfillRequest`, so nothing on the webroot moved), and A/B'd against the deployed file — which was confirmed byte-identical to `origin/working-1.6` first, so no branch drift could be mistaken for an effect.

| grid | rows | before | after | rendered widths |
|---|---|---|---|---|
| host | 86 | 870ms | **571ms** | identical |
| image | 29 | 194ms | **108ms** | identical |
| snapin | 4 | 39ms | 25ms | identical |
| group | 2 | 23ms | 14ms | identical |

Header and body column widths and the Responsive collapse set are byte-identical in every arm. Both `phpstan` passes clean.

`FOG_BCACHE_VER` bumped 339 → 340, since `fog.common.js` changed.

## What this does not fix

With `responsive: false` the cost is **0ms**, but two columns stop collapsing and the table overflows horizontally — so Responsive is genuinely earning its keep here, not dead weight. This change recovers the redundant third of its cost; the rest is the feature itself, and whether wide desktop grids should opt out of it is a UX decision, not a perf cleanup.

## Downstream

None — no route or class-list change.

### DIFF
--- a/packages/web/management/js/fog/fog.common.js
+++ b/packages/web/management/js/fog/fog.common.js
@@ -3222,6 +3222,100 @@ function fogFetchIfShort(dt, waited) {
   }
   dt.draw(false);
 }
+/**
+ * Stop DataTables Responsive re-measuring every loaded row on every adjust.
+ *
+ * Responsive answers `column-sizing` -- which every columns.adjust() fires --
+ * by running _resizeAuto(), and _resizeAuto() deep-clones the header, the
+ * footer and EVERY row of the current page into a hidden table so the browser
+ * will compute each column's natural width. With Scroller the "current page"
+ * is every row fetched so far, so the cost is linear in the size of the list.
+ * Measured on a 86-host list: 8 calls per load at ~140ms each, of which only
+ * 15ms is building the clone -- the other ~125ms is the layout the browser is
+ * forced into to answer offsetWidth.
+ *
+ * Most of those calls cannot produce a new answer. The widths _resizeAuto()
+ * reads come only from the markup it clones; it measures with the table at
+ * width:auto inside a 1px hidden box, so the viewport is not an input. Five of
+ * the eight calls returned a vector byte-identical to the call before them.
+ *
+ * So sign exactly what gets cloned -- the table node's class list, the visible
+ * column set, and the header/body/footer HTML -- and when the signature is
+ * unchanged, restore the previous answer instead of re-measuring. Signing the
+ * markup rather than a draw counter is what keeps this correct: Responsive
+ * converges over two passes, adding `dtr-control` to the first column between
+ * them, and that genuinely does change the answer. A coarser key would cache
+ * the pre-convergence widths and leave the column permanently 17px narrow.
+ *
+ * The class list is sorted before signing because addClass()/removeClass()
+ * reorder the tokens without changing what renders, and that alone defeated
+ * the cache twice per load.
+ *
+ * The signature costs ~1.5ms against the ~140ms it skips. Nothing else caches
+ * it, so a stale entry cannot outlive the table.
+ *
+ * Dropped on window resize: the markup does not change when the browser zooms
+ * or the font size does, but the metrics do.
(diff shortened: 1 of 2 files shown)


mastacontrola:
Correction to the PR body, for the record: **"A 1000-host list implies ~10s" is wrong.** The cost is bounded, not unbounded.

Scroller derives its page length from the viewport (`viewportRows * displayBuffer` — 144 on a 1600×1000 window) and virtualises, so `rows({page:'current'})` never exceeds that however long the list is. The ceiling is therefore ~144–180 rows, i.e. ~230ms per `_resizeAuto` call and ~1.8s a load on a large list — worse than the 86-row case measured here, but roughly 2×, not 12×.

What *does* scale it directly is **`FOG_VIEW_DEFAULT_SCREEN`**: page length feeds straight into how many rows get cloned, so raising it multiplies this cost one-for-one.

Two other numbers worth recording, measured while checking the above:

- The clone-and-layout half alone is only ~**0.2ms/row** and flattens with size (10 rows 0.9ms/row, 1000 rows 0.19ms/row). There is a fixed floor of ~8–9ms per call, which is why a 4-row grid still costs 39ms across 7 calls — that is per-call overhead, not per-row work.
- Most of the rest is **DataTables API iteration**, not layout: `rows().every()` / `cells().every()` plus 946 individual jQuery cell clones at 86 rows.

None of this changes the fix or the measured before/after — the redundant calls were redundant either way.

PR #1523, opened 2026-08-30
Requeue a merge queue entry whose checks were never dispatched

mastacontrola:
Requeue a merge queue entry whose checks were never dispatched

PR #1522 was enqueued at 17:59:00Z. GitHub created the gh-readonly-queue
branch and then never dispatched a run against it:

    gh api repos/.../actions/runs?head_sha=<queue sha> --jq .total_count
    0

Still 0 fifteen minutes later, while the four entries before it that hour
had each dispatched within seconds. The workflow file ON THE QUEUE REF
carried merge_group:, the queue base was the current tip of working-1.6,
and nothing here has a concurrency group that could have cancelled it.
No GitHub incident was open. It was a dropped dispatch -- their
2026-08-26 incident, "2.6% of workflow runs experienced delays", is the
same class of thing when it is large enough to publish.

Dequeuing and re-arming auto-merge by hand built a fresh merge group that
dispatched in six seconds and merged. This automates that one repair.

WHY IT NEEDS A WATCHDOG RATHER THAN VIGILANCE

Nothing about it looks wrong. The pull request's own checks are green and
that is what the checks list shows. The entry reports AWAITING_CHECKS,
which is what it would report if the checks were merely slow. The missing
thing is a run that does not exist, so there is no red mark, no
annotation and no notification. The only signal is an absence.

Left alone it does not sit orange forever, which is worse. The ruleset's
check_response_timeout_minutes is 60, so on the hour the entry is ejected
and the pull request goes back to open, green and CLEAN -- which reads as
finished. A dequeue disarms auto-merge (observed directly: dequeuing
#1522 by hand left autoMergeRequest null), and automerge.yml re-arms only
on opened and ready_for_review. So nothing puts it back, and the pull
request lands in exactly the state that file exists to prevent, by way of
the mechanism meant to prevent it.

WHAT IT DOES

An entry qualifies only when it has been AWAITING_CHECKS longer than
STALL_MINUTES *and* its queue commit has neither a check run nor a commit
status -- not "slow", but "nothing was ever dispatched". Both are read,
because Actions reports check runs while older integrations report
statuses, and reading one alone would requeue an entry that is running
perfectly well.

Every 15 minutes against a 15-minute threshold puts worst-case detection
at 30 minutes, leaving two further attempts before the 60-minute
eviction.

DELIBERATELY NOT INCLUDED

Re-arming auto-merge on a pull request that is green, open and unarmed.
That would also catch the post-eviction state, but it would fight
`gh pr merge --disable-auto`, which automerge.yml documents as the way to
hold a single pull request back; and whether an eviction disarms
auto-merge the same way a manual dequeue does is untested. Catching the
stall inside 15 minutes should mean the eviction never arrives. If one
ever does, that is the point to design the second arm against something
observed rather than assumed.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/queue-watchdog.yml
+++ b/.github/workflows/queue-watchdog.yml
@@ -1,0 +1,164 @@
+name: Merge queue watchdog
+
+# Requeues a merge queue entry whose checks GitHub never dispatched.
+#
+# WHAT GOES WRONG
+#
+# Enqueuing a pull request makes GitHub build a throwaway `gh-readonly-queue`
+# branch -- base plus the pull request -- and run the required checks against
+# THAT commit, not against the pull request's own head. That is the whole
+# point of a queue: the checks that decide the merge are the checks on the
+# merge result.
+#
+# On 2026-08-30, pull request #1522 was enqueued at 17:59:00Z. The queue
+# branch was created. No workflow run was ever dispatched against it:
+#
+#     gh api repos/.../actions/runs?head_sha=<queue sha> --jq .total_count
+#     0
+#
+# Fifteen minutes later it was still 0, while the four entries before it that
+# hour had each dispatched within seconds. The workflow file ON THE QUEUE REF
+# carried `merge_group:`, the queue base was the current tip, and nothing in
+# this repository has a concurrency group that could have cancelled it. There
+# was no GitHub incident open. It was a dropped dispatch, below the threshold
+# they publish -- their 2026-08-26 incident, "2.6% of workflow runs
+# experienced delays", is the same class of thing when it is big enough to
+# report.
+#
+# WHY IT NEEDS A WATCHDOG AND NOT JUST VIGILANCE
+#
+# Nothing about this looks wrong. The pull request's OWN checks are green and
+# that is what the checks list shows; the entry reports AWAITING_CHECKS, which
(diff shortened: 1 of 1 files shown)


mastacontrola:
Closing: the diagnosis in this commit is wrong.

This is not a dropped dispatch. It is deterministic, and #1523 reproduced it while I was still writing the watchdog for it.

GitHub will not start a workflow run from an event caused by GITHUB_TOKEN. automerge.yml arms auto-merge with secrets.GITHUB_TOKEN, the enqueue is attributed to whoever armed it, so the merge_group event raised by a bot-armed enqueue starts nothing at all -- including tests.yml, the only workflow that supplies the eight required contexts.

  PR    enqueued by             merge_group run
  1519  mastacontrola           dispatched
  1520  mastacontrola           dispatched
  1521  mastacontrola           dispatched
  1522  github-actions[bot]     NONE
  1522  mastacontrola (requeue) dispatched, 6s later
  1523  github-actions[bot]     NONE

Superseded by the App-token fix, which removes the condition rather than polling for it. A scheduled bot with contents:write manipulating the merge queue is not worth carrying to work around a config bug that has an actual fix.

PR #1544, opened 2026-08-31
Fix the boot fatal and the blind gate that let it through

mastacontrola:
Follow-up to #1540, found by installing v1.6.23 into a worktree rather than by any run of the suite.

`tests/getclass-resolves-without-aliases.test.php` collected a plugin's declarations by matching the pre-1.6 discovery suffixes — `.class.php`, `.page.php` and friends. ADR 0035 took those away, so on a tree that actually has plugins the loop finds nothing and all **43** plugin `getClass()` literals report as resolving through an alias that no longer exists:

```
FAIL:
  - getClass() is called with 43 name(s) that resolve through nothing this tree
    declares ...: Capone, CaponeManager, HelloWorld, ..., WindowsKeyManager, location
```

**The code was right the whole time.** `qualify()` answers a short name from the derived path, so `getClass('LDAPManager')` reaches `FOG\Plugins\LDAP\Managers\LDAPManager` — verified live against the real v1.6.23 tree. Only the gate was blind.

## Why it survived the layout change

`packages/web/lib/plugins` is *fetched*, not tracked. It is absent in CI, in a fresh clone, and in every worktree I verified #1540 in — so the plugin loop found no files, every plugin literal fell through to the `class_exists()` escape, and the check passed **by never running that arm at all**. It goes red only on a tree with plugins actually installed, which is to say on a real server and nowhere else. That is the same failure shape the repo already warns about: a gate that passes for reasons the real command never touches.

## The fix

Collect from the layout that exists — `<plugin>/src/<Bucket>/<Class>.php`, keyed on the basename, which is what a `getClass()` literal holds.

Both directions mutation-verified against the installed v1.6.23 tree:

- delete `ldap/src/Managers/LDAPManager.php` → red, naming `LDAPManager`
- rename it so the class no longer matches its file → red, naming `LDAPManager`
- restored → green, 146 literals resolve

The `lib/{pages,hooks,reports,events}` collector above it keeps its suffix match deliberately and gains a comment saying why: all 52 of those classes moved to `src/`, so it finds nothing on a current tree — but a webroot upgraded rather than laid fresh can still be carrying them, and a name that resolves only there is exactly what this file exists to name.

Test-only change. `phpstan-tests.neon` clean.

### DIFF
--- a/tests/getclass-resolves-without-aliases.test.php
+++ b/tests/getclass-resolves-without-aliases.test.php
@@ -199,7 +199,13 @@ foreach ($dirs as $dir) {
     }
 }
 
-/** The 46 discovery-named classes under lib/, which carry their own aliases. */
+/**
+ * The discovery-named classes that used to live under lib/{pages,hooks,
+ * reports,events}. All 52 moved to src/<Bucket>/<Class>.php, so this finds
+ * nothing on a current tree -- it is kept because a webroot upgraded rather
+ * than laid fresh can still be carrying them, and a name that resolves there
+ * and nowhere else is exactly what this file exists to name.
+ */
 $libClasses = [];
 foreach (['pages', 'hooks', 'reports', 'events'] as $dir) {
     foreach (glob($web . '/lib/' . $dir . '/*.php') as $file) {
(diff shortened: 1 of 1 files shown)


mastacontrola:
Second commit added — this PR now carries the **blank-500 fix**, which is the urgent half.

`class Config` is generated by the installer into `commons/config.class.php`. It is the one `.class.php` file left anywhere in a deployed tree, and the only class in FOG that is neither PSR-4 under `src/` nor inside a plugin — so it was reached by the autoloader's recursive `*.class.php` scan of BASEPATH, and that scan is exactly what #1540 retired. `new Config()` in `startInit()` became:

```
Uncaught Error: Class "Config" not found in commons/init.php:1383
```

on **every page**. A blank white page in a browser, and `installfog.sh`'s own "Checking web server serves FOG" step failing with `curl exit 22, 0 bytes of body`. Found by installing the merged tree; no run of the suite could see it.

It is now required by name, the same way `fogpaths.php` already is in `_loadPaths()` — a generated file at a fixed path gets pulled in, not found. A missing one names itself and says to re-run the installer, instead of dying three lines later on a class nobody can locate.

`config.class.php` was the *only* casualty: `find` over a deployed webroot returns exactly one `.class.php` outside `lib/plugins`, and it has exactly one reference in the tree.

## Why nothing caught it

Every PHP test in the suite constructs `Initiator` — which only registers the autoloader — and stops before `startInit()`, because `startInit()` wants a database. The whole boot tail had no coverage at all. That is the same shape as the other commit here: a check that passes because the arm under test never runs.

`tests/generated-config-is-loaded.test.sh` closes it without a database, by driving `commons/base.inc.php` in a tree whose `config.class.php` is a stub and asserting on *which* error comes back — the guard fires and names the file when it is absent, and the boot reaches `PDODB` when it is present, which is the proof that `Config` resolved.

It boots through `base.inc.php` rather than `management/index.php` deliberately: `index.php` reaches its includes as `../commons/...`, which PHP resolves against the **real** path, so a fixture that symlinks `management/` runs against the repository's own `commons/` and never sees the stub. My first cut did exactly that and passed for the wrong reason.

Mutation-verified four ways, all red — require deleted; require pointed at a misspelled filename; require moved after the use; `is_readable` guard inverted. The last three are what a textual check for "`require_once` near `new Config()`" would have waved through.

Full suite 244/0, both phpstan passes clean.

PR #1592, opened 2026-09-01
Make the Certificates page a table rather than a train of cards

mastacontrola:
The Certificates page showed the root's subject and SHA-256 as prose in one card, the imported root's subject, SHA-256 and expiry as prose in a second, and the remaining six slots in a table carrying neither — the same three facts in two shapes, six certificates with no fingerprint anywhere on the page, and 2400px of cards between any two rows you wanted to compare.

Every certificate is now one row of one table.

| | before | after |
|---|---|---|
| Slots showing a fingerprint | 2 of 8 | 8 of 8 |
| Cards | 5 | 4, one of them collapsed |
| What signed a certificate | not shown | sub-line under the subject |
| Expiry status | date only | date, plus red past / amber inside 30 days |

## What changed

- **One table**: label and purpose, subject with its issuer, expiry, SHA-256, download. The fingerprint is the reason anyone opens this page — it is what you compare against a client's trust store when a client stops authenticating — and it reached two of the eight slots.
- **Expiry badges.** Red once past, amber inside 30 days, which is the window every ACME client renews in. A certificate still amber there is one nothing is renewing. Rendered with `gmdate()`: openssl's `notAfter` is already GMT, and relabeling it into `FOG_TZ_INFO`'s zone — the zone rows are *stored* in — is how dates elsewhere have named the wrong day.
- **Install preferences** become a row each (switch, setting, what it does) instead of three stacked prose blocks.
- **"Using your own PKI"** — reference material for a migration nobody does twice, and the longest thing on the page — arrives collapsed.
- **External root CA** keeps only its control; what is imported is now a row in the table above, so the card stopped repeating the subject, fingerprint and expiry a third time.

Two supporting changes, both confined to this page:

- `_box()` takes a `cardClass` option so a card can render already collapsed. AdminLTE hides a `.collapsed-card`'s body in CSS, so the class in the markup is the whole mechanism.
- its collapse toggle emits both `data-lte-icon` elements. `FOGPage`'s shared `$FOGCollapseBox` is a hardcoded `fa-minus`, so a card rendered collapsed sat there offering to collapse itself again.

No JavaScript change — the switches keep their ids, `.pki-pref` class and `data-action`, and the downloads are still plain links. No CSS change, so `FOG_BCACHE_VER` is untouched.

## Verification

- Rendered against a shadow tree with an eight-slot PKI fixture, at 1440px and 900px, in both themes: no horizontal overflow, tables scroll inside their own `.table-responsive`.
- The helper-absent path (`_pkiStatus()` returns null) still renders the warning and nothing else, and the exposed-private-key danger card still leads the page.
- `tests/certificate-table.test.php` (new, 21 checks) renders the table and asserts every present slot is a row carrying its fingerprint, an absent slot is not a row, a self-signed certificate says so rather than naming itself as issuer, and the badge thresholds. The UTC check runs from `America/Los_Angeles` against a `notAfter` half an hour after midnight, so `date()` and `gmdate()` land on different calendar days and only one can pass.
- **Every assertion was made to fail** against a mutation of the code it covers: `gmdate`→`date`, the 30-day branch removed, the fingerprint cell emptied, the present-slot skip removed, the issuer sub-line dropped, `collapsed-card` removed, and the two-icon toggle reverted.
- Suite: 270 passed, 0 failed. Both `phpstan` passes clean.

Nothing here changes what the page can do or who may do it — `certificates()` is still `settings.view`, the POST actions are still `system.pki`, and `tests/certificate-management-permission.test.php` passes unchanged.

https://claude.ai/code/session_01CreCNucEFLNcrzbZCjchvJ

### DIFF
--- a/packages/web/management/languages/de_DE.UTF-8/LC_MESSAGES/messages.po
+++ b/packages/web/management/languages/de_DE.UTF-8/LC_MESSAGES/messages.po
@@ -115,6 +115,10 @@ msgid "%d certificates in this file"
 msgstr ""
 
 #, php-format
+msgid "%d days left"
+msgstr ""
+
+#, php-format
 msgid "%d host(s) are assigned an image they cannot run"
 msgstr ""
 
@@ -2159,10 +2163,6 @@ msgstr "Aktuelle Datensätze"
 msgid "Current member-host AD state"
 msgstr ""
 
-#, fuzzy
-msgid "Currently imported"
-msgstr "Aktuelle Datensätze"
-
 msgid "DMI Field"
 msgstr ""
 
@@ -2528,7 +2528,7 @@ msgstr "Dauer"
 msgid "Each client needs this certificate enrolled once, by someone physically at the machine"
 msgstr ""
 
-msgid "Each of these is a public certificate. Downloading one is how you add this server to another machine's trust store, hand an auditor the chain, or check what a client is actually being offered."
+msgid "Each of these is a public certificate. Downloading one is how you add this server to another machine's trust store, hand an auditor the chain, or check what a client is actually being offered; the SHA-256 is what to compare against a client's trust store when working out why one stopped authenticating. Private keys are deliberately absent -- nothing on this page can read one, and nothing on it can hand one out."
 msgstr ""
 
 msgid "Edit"
@@ -2731,6 +2731,10 @@ msgstr "Bestehendes Objekt muss boolean sein"
 
 msgid "Expand - walk the chain (any directory)"
 msgstr ""
+
+#, fuzzy
+msgid "Expired"
+msgstr "Sitzungsname"
 
 msgid "Expires"
 msgstr ""
@@ -4753,6 +4757,10 @@ msgstr "Token zurücksetzen"
 msgid "Issue a token on behalf of a user"
 msgstr ""
 
+#, php-format
+msgid "Issued by %s"
+msgstr ""
+
 msgid "Issuer"
 msgstr ""
 
@@ -6285,6 +6293,9 @@ msgstr ""
 msgid "One entry per row. The element shape follows the requested fields: a scalar when a single column is projected, an object when several are."
 msgstr ""
 
+msgid "One is imported. It is the \"Imported root\" row in the table above."
+msgstr ""
+
 msgid "One or more files must be uploaded via the \"snapinfiles[]\" multipart field"
 msgstr ""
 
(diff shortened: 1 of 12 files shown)


mastacontrola:
## Reworked after review

Two things were wrong with the first cut.

**It wasn't the house pattern.** Nearly every management page in FOG is one card with a tab strip built by `FOGPage::tabFields()`; this was five stacked cards. It is now four tabs through that same helper — Certificates, External root CA, Install preferences, Using your own PKI.

**The collapsed card could not be opened.** AdminLTE binds the collapse toggle with `document.querySelectorAll(...).forEach(el => el.addEventListener(...))` once at `DOMContentLoaded`. It is **not delegated**, and FOG replaces the content by AJAX on every sidebar click — so a card arriving by navigation has no handler on its toggle at all. "Using your own PKI" shipped rendered `collapsed-card`, which made it content nobody could reach.

It works only on a hard reload — which is exactly what the static snapshot I verified it against was. The check passed for a reason the real navigation never touches. This time it was driven the way you actually use it: dashboard → click Certificates in the sidebar → click each of the four tabs, and every pane opens. Bootstrap's own tab toggle *is* delegated, which is why the house pattern works where that did not.

So `_box()`'s `cardClass` option and its two-icon collapse toggle are reverted — nothing needs them now, and neither could ever have worked.

### Also in this push

- **The alarms stay outside the tab card, above it** — an exposed CA private key, and a missing PKI helper. A condition an administrator must act on cannot live behind a tab they have to know to click.
- The five section builders **return** their body instead of echoing a card, which is what lets `certificates()` drop a tab whose content is empty rather than render an empty pane. On a storage node, with no helper, three of the four have nothing to say.
- `_box()` emits no `card-header` when there is no title, help or collapse button — a card inside a tab pane takes its title from the tab.
- `FOGPage::tabFields()` gains `bool` in its `@param` for `$obj`. Its own docblock already said "Anything falsy means there is no object" and it branches on `if ($obj)`; the declared `int|object` was simply wrong, and two of the three callers passing `false` carried a baselined phpstan error for saying so. **Both baseline entries are retired rather than a third being added.**

### Gate

`tests/certificate-table.test.php` (33 checks) now also pins that the page is a tab card, that an empty section is not an empty tab, that the alarms render above the tabs, and that **nothing on the page ships behind a collapse an AJAX visit cannot open** — asserted on the rendered page, so re-introducing it any other way still fails. It gains coverage of the key-exposure alarm's three branches, including the `expect_readable` one that stops a correct install reporting itself as a breach.

Every assertion was made to fail against a mutation of the code it covers. One gap I'll name rather than paper over: deleting the `_certificateKeyExposure()` call from `certificates()` outright is still not caught, because `_pkiStatus()` caches in a function-static and shells out to a root-only helper, so a test cannot force a non-null status through the page. The builder itself is covered in all three branches, and "alarms precede the tab card" is gated.

Suite 270/0, both phpstan passes clean, PSR-2 clean.

PR #1595, opened 2026-09-01
Set error_log, and re-exec updatefog.sh from a copy before it rewrites itself (GH-1594)

### LINKED ISSUE #1594: updatefog.sh runs with $error_log unset: every redirect writes to a file named ""
## Summary

`bin/updatefog.sh` reads `$error_log` in four places and never sets it. The
assignment was lost when the script was rewritten in #1584.

```bash
$ grep -n 'error_log' bin/updatefog.sh
240:    echo " * Git update failed -- nothing was installed. See $error_log."
265:(cd "${FOG_git_path}/bin" && bash installfog.sh ... 2>&1 | tee -a "$error_log"; ...)
275:# is one worth naming (offerRevert) -- but it printed that into $error_log,
278:echo " * The system has NOT been reverted. See $error_log -- if the checkout"
```

There is no `error_log=` line.

## Impact

- `gitUpdateToBranch()` (`lib/common/update.sh`) redirects every git command
  with `>>$error_log 2>&1`. Unset, that is a redirect into a file named `""`,
  so each one fails before the git command runs and the update cannot proceed.
- `tee -a "$error_log"` fails the same way, so the installer's output is not
  captured.
- The failure messages read "See ." with no path.

`errorStat` is not the mitigation here: `$exitFail` is set in this script
precisely so a failed git step returns rather than exiting, so the run
continues past the broken redirects.

## Fix

Restore the assignment, after the working directory is resolved and before
anything writes to it:

```bash
[[ ! -d ./error_logs/ ]] && mkdir -p ./error_logs >/dev/null 2>&1
error_log="${workingdir}/error_logs/fog_update_error.log"
: > "$error_log"
```

## Note

This is a regression I introduced in #1584 and did not catch, because the test
added there exercises `offerRevert()` and `markInstallCommit()` directly rather
than running `updatefog.sh` end to end. The fix adds assertions that
`error_log` is set and set before its first use.

darksidemilk:
Fixes #1594. **Draft.**

Two fixes to `bin/updatefog.sh` — one a regression I introduced in #1584, one a latent bug the regression fix made obvious.

## `$error_log` is never set (GH-1594, live on working-1.6 now)

The assignment was lost when the script was rewritten. It's read four times and set nowhere.

`gitUpdateToBranch()` redirects **every** git command with `>>$error_log 2>&1` — unset, that's a redirect into a file named `""`, so each one fails before the git command runs. The `tee` that shows the installer fails the same way, and the failure messages read *"See ."* with no path. `exitFail` is set here on purpose, so the run continues past the broken redirects instead of stopping at the first.

My test in #1584 didn't catch it because it exercises `offerRevert()` and `markInstallCommit()` directly rather than running the script end to end. It now asserts `error_log` is set, and set *before its first use*.

## The script replaces itself

`gitUpdateToBranch()` checks out a different branch, which rewrites `bin/updatefog.sh` underneath a bash process still reading it. Bash reads a script incrementally and seeks by byte offset, so a file that changes length mid-run resumes in the middle of a different line — silently, arbitrarily, and only **after** the checkout has succeeded: the code has moved and the process driving the update is executing fragments. Every channel switch changes this file.

So it copies itself out of the checkout and re-execs before going near git. `error_log` is set after that, so it lands in the checkout rather than beside the temporary copy.

## What the test proves, and what it doesn't

Being straight about this. Bash reads ahead in blocks, and this script is small enough that a run can survive the truncation **without** the copy — the fixture was observed doing exactly that. So the behavioural case is a regression check (the guard must not break the ordinary path), **not** proof the guard is required, and the comment says so rather than letting a future reader take it for more than it is.

What pins the guard is the pair of structural assertions: that the re-exec exists, and that it comes **before** `gitUpdateToBranch`. Those do fail when the guard is removed — verified.

`tests/revert-offer.test.sh`: 30 assertions, all passing.

## Related

The same guard is already in the 1.5 port (#1589), where the hazard is worse — that file doesn't exist on the branch being left and appears only after the checkout.

https://claude.ai/code/session_01XtZHpyWZWDPSxePdTViGZy

### DIFF
--- a/bin/updatefog.sh
+++ b/bin/updatefog.sh
@@ -90,6 +90,9 @@ optargs=$(getopt -o $shortopts -l $longopts -n "$0" -- "$@")
 [[ $? -ne 0 ]] && usage
 eval set -- "$optargs"
 
+# Kept before the option loop shifts them away: the re-exec below needs the
+# ORIGINAL arguments.
+origArgs=("$@")
 autoYes=""
 while :; do
     case $1 in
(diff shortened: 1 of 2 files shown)


mastacontrola:
Pushed two commits here while reviewing:

**`8a3c60ed3`** — US spelling (`BEHAVIOURAL`/`behavioural` → `BEHAVIORAL`/`behavioral`) in the comment block added to `tests/revert-offer.test.sh`. That was the only failing check; no assertion changed. Suite is 276/276 locally.

**`83ff32b78`** — part 2 of #1597. The failure block named `$error_log` without showing it, so an admin had to go and open the file to see the git error. It now tails the log inline, which is what `errorStat`'s own non-`exitFail` path already does:

```
 * Git update failed -- nothing was installed.
 |
 | error: pathspec 'working-1.6' did not match any file(s) known to git
 |
 * Full log: /root/fogproject/error_logs/fog_update_error.log
```

This landed here rather than in its own PR because it depends on the `error_log` assignment this PR adds — before it, that line renders as `See .` (GH-1594).

Also confirmed GH-1594 is live on `working-1.6` right now, by running the real code rather than reading it: nothing in `updatefog.sh`'s sourcing chain (`functions.sh`, `config.sh`, `update.sh`) assigns `error_log`, and `>>$error_log` unset is an `ambiguous redirect` — exit 1, git never runs. So `gitUpdateToBranch` fails on its first command every time:

```
 * Fetching FOG (working-1.6)........ambiguous redirect
Failed!
 * Git update failed -- nothing was installed. See .
```

With this PR's fix the same harness fetches, and the real git error reaches the log.

One separate finding, filed as #1597 with its own PR (#1599): `errorStat` prints `Failed!` and then `OK` for the same step whenever `exitFail` is set — visible in the trace above. Pre-existing in `functions.sh`, not from this PR, but it affects all four scripts that set `exitFail`.

PR #1606, opened 2026-09-01
Keep AdminLTE's card tools working after an AJAX page load

mastacontrola:
Follows #1600, which found this. **First, a correction to what I told you there:** I said every collapsible card in the UI was half-broken. That overstated it — the honest scope is below.

## The bug

adminlte4 binds collapse, remove and maximize exactly once, inside its own `DOMContentLoaded`:

```js
document.querySelectorAll('[data-lte-toggle="card-collapse"]')
        .forEach(el => el.addEventListener('click', …))
```

Not delegated. `doPageLoad()` replaces `#ajaxPageWrapper` wholesale on every sidebar click, so a card arriving by navigation has no handler on its toggle.

## Actual scope: three buttons on one page

I checked every render site rather than assuming:

| Site | Reached by | Status |
|---|---|---|
| `ClientManagement` (`node=client`) | **sidebar link → AJAX** | **broken today** — 2 collapse buttons dead |
| `ImageManagement` edit | grid link → full page load | fine, AdminLTE binds it |
| `ServerInfo` (`node=hwinfo`) | not in the sidebar | fine |
| `_box(collapse => true)` | no caller passes it | n/a |

It went unreported because a card that starts *open* only degrades to "cannot be collapsed", which reads as a feature nobody uses. A card rendered `collapsed-card` is the loud version — its body is `display:none` from AdminLTE's own CSS, and the only thing that removes the class is the listener that was never attached. That is how #1600 found it.

## The fix

One delegated listener on `document`, in the **capture** phase. Capture does two jobs:

- it runs before the element's own listener whether or not AdminLTE attached one, and
- `stopPropagation()` there keeps the event from reaching the target — so a card that *was* present at `DOMContentLoaded` isn't toggled twice by two handlers, which would collapse and re-expand in one click and look just as dead.

That also means the fix doesn't depend on script order or on which cards were present when. There is already precedent for this exact shape in `fog.common.js` (the DataTables collection keydown suppressor).

Not re-running AdminLTE's initializer per page load: it isn't exported, and re-running it would re-bind the persistent chrome outside `#ajaxPageWrapper` once per navigation.

`FOG_BCACHE_VER` 355 → 356, since shipped JS changed.

## Verification — in a browser, on the route that failed

Shadow tree, real app, real AJAX navigation. Not a static snapshot; that is what let the #1592 bug through.

| Case | Result |
|---|---|
| Sidebar → Client, click collapse | collapses, then re-expands |
| **Same page and route, this block deleted** | **click does nothing** — bug reproduced |
| Same page by full page load (both handlers exist) | toggles once per click — no double-fire |
| Remove / maximize on a card injected after load | both work, including a click landing on the child `<i>` |

## Gate

`tests/card-tools-survive-ajax-nav.test.php` (14 checks) pins the **contract with the vendored library** rather than restating this code: every tool AdminLTE binds is one FOG takes over, and every `CardWidget` method FOG names still exists on the class — both read out of `adminlte4.min.js`. An upgrade that adds a fourth tool or renames a method fails the build instead of shipping a dead button.

It asserts on the code **with comments stripped**. `stopPropagation`, the capture flag and all three tool names also appear in the docblock above the handler, and two checks were passing against that prose until a mutation run caught them — so the mutation pass earned its keep here.

Mutations proven to fail: capture flag dropped, `stopPropagation` dropped, a tool no longer taken over, a `CardWidget` method renamed to one that doesn't exist, the whole takeover removed, and `FOG_BCACHE_VER` left behind.

What the suite **cannot** prove, since it has no JS engine: that the click actually works. That is the browser table above.

Suite 279/0, both phpstan passes clean, PSR-2 clean.

https://claude.ai/code/session_01CreCNucEFLNcrzbZCjchvJ

### DIFF
--- a/packages/web/management/js/fog/fog.common.js
+++ b/packages/web/management/js/fog/fog.common.js
@@ -4592,6 +4592,73 @@ function macVendorIcon(vendor) {
 })();
 
 /**
+ * AdminLTE's card tools, re-armed for FOG's AJAX navigation.
+ *
+ * adminlte4 binds them exactly once, inside its own DOMContentLoaded:
+ *
+ *   document.querySelectorAll('[data-lte-toggle="card-collapse"]')
+ *           .forEach(el => el.addEventListener('click', ...))
+ *
+ * That is not delegated, and doPageLoad() replaces #ajaxPageWrapper wholesale
+ * on every sidebar click -- so every card that arrives by navigation, which is
+ * almost every card anyone ever sees, has a dead toggle. The three tools are
+ * all bound that way: collapse, remove and maximize.
+ *
+ * It went unreported for as long as it did because a card that starts OPEN
+ * only degrades to "cannot be collapsed", which reads as a UI nobody uses. A
+ * card rendered `collapsed-card` is the loud version: its body is
+ * display:none from AdminLTE's own CSS and the only thing that removes the
+ * class is the listener that was never attached, so the content is
+ * unreachable. That is how it was found (GH-1600).
+ *
+ * CAPTURE phase, deliberately. The handler has to run before the element's own
+ * listener whether or not AdminLTE managed to attach one, because a card that
+ * WAS present at DOMContentLoaded still carries it -- and two handlers both
+ * calling toggle() would collapse and expand in the same click, leaving the
+ * button looking just as dead. stopPropagation() during capture keeps the
+ * event from reaching the target at all, so exactly one of the two runs and it
+ * is this one. That also means the fix does not depend on script order, on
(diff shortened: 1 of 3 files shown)


mastacontrola:
Closing in favor of a fresh PR from the same branch: this one never fired its `pull_request` event (zero workflow runs across an open, a close/reopen and a force-push), so no checks could ever report. PR #1608 was opened in the same minute and is in the same state, so it looks like a GitHub-side event-delivery gap for that window rather than anything about the branch.

PR #1608, opened 2026-09-01
Remove the ClamAV virus scan (GH-328)

### LINKED ISSUE #328: Remove Virusscan option
Details in the forum topic: https://forums.fogproject.org/topic/13518/fog-1-5-6-arch-linux-antivirus-scan

Cross linking this to the fos issue #28 as there are components in both repos: https://github.com/FOGProject/fos/issues/28

mastacontrola:
Closes #328 (needs closing by hand — closing keywords are inert off the default branch).
FOS half is FOGProject/fos#28; see **Ordering** below.

## The feature was already broken on this branch

Worth stating first, because it changes what this PR is:

| | |
|---|---|
| FOS `bin/fog.av` posts its findings to | `service/av.php` |
| That endpoint exists in | `/var/www/html/fog-1.5/service/av.php` |
| …and in `packages/web/service/` on this branch | **it does not** |

So on 1.6 a virus scan boots, mounts the NFS ClamAV share, updates its definitions, scans the Windows partition, and throws every result away against a 404. Nothing writes the `virus` table on 1.6, and nothing reads it either — there is no `Virus` model, manager, report or page anywhere in the tree, only the rows a 1.5 install left behind. What was still on offer was two task types (21 and 22) for a scan that could not report, plus a `clamav=` kernel argument feeding a boot image with nowhere to send its output.

This finishes the removal rather than starting one.

## What changes

* **Schema step 406** deletes task types 9, 21 and 22 and drops the `virus` table.
* **No menu code changes.** Every task list is built from the `taskTypes` table itself — `FOGPageRender::taskTypeAccordion()` through `Route::getList('tasktype')` — so the rows *are* the feature. Removing them clears the host page, the group page, the Queue Task modal and the API in one move.
* `BootMenuBase` no longer builds a `clamav=` kernel argument or special-cases task types 21/22.
* `virus` joins `imagingLog` in the manifest's `retired` block; its `schema-constraints.php` row goes.
* Two counts in `docs/adr/0031-…` and `docs/development/foreign-keys.md` move 121 → 120 (polymorphic subtotal 7 → 6), which is what `tests/foreign-key-map.test.php` checks.

The historical steps that *created* the table and seeded the rows are untouched — `schema.php` is an append-only replay log.

## The delete order is the whole risk

`tasks.taskTypeID` and `scheduledTasks.stTaskTypeID` both reference `taskTypes.ttID` **ON DELETE RESTRICT** (ADR 0031, groups 6 and 5), and 1451 is not in `Schema::runSteps()`'s skippable-error list.

A step that deletes the parent first therefore **aborts the schema update on `?node=schema`** — but only on a server that has actually queued a scan. A clean test database has none, so that failure is invisible to every static check and to the whole suite. Children first, parent last.

## What is lost, said plainly

A **scheduled** scan can never run again, so deleting it removes nothing usable.

A **`tasks`** row is history, and this destroys it. The run itself survives where the project already decided run history lives: `taskLog` carries `logTaskTypeName` as text, takes no constraint at all (ADR 0021), and is written by `TaskLog::recordState()` on every state change. That is the same argument ADR 0022 decision 3 used to retire `imagingLog`. The step records what it deleted through `error_log()` and `Audit::record()` rather than swallowing it.

Task type **9** is the original 1.x row; step 33 truncated `taskTypes` and reseeded without it, so no current server has one. It is named in the delete anyway, for a database restored from a pre-33 dump.

## Verification — against a real database, not just the source

`background_scripts/prove_schema_406_virus_removal.php` runs against a clone of the live 1.6 database in a throwaway MariaDB container, with a fixture scan task, a scheduled scan and a `taskLog` row inserted first so the RESTRICT is armed:

```
--- 1. the RESTRICT constraints are actually present ---
ok   taskTypes has RESTRICT children in this database (found 2)
--- 2. the naive order aborts at 1451 ---
ok   deleting the task types first is refused, errno 1451 (got 1451)
ok   and the task types are still there afterwards
--- 3. step 406 runs clean ---
FOG schema 406: removed the ClamAV virus scan (GH-328). 1 task(s) and 1 scheduled
task(s) referencing task types 9, 21 and 22 were deleted so the task types
themselves could be. The runs remain in taskLog.
ok   step 406 returned no error (NULL)
--- 4. the end state ---
ok   no virus scan task types remain
ok   no tasks reference them
ok   no scheduled tasks reference them
ok   the virus table is gone
ok   the taskLog row for the scan survives, naming the task type
ok   nothing else was deleted from tasks
--- 5. idempotent ---
ok   a second run returns no error (NULL)
PASS
```

`tests/virus-scan-is-removed.test.php` pins the ordering, the retirement and the boot-menu removal — and checks that **every enabled RESTRICT child of `taskTypes`** is one the step clears, so a constraint added later without a matching delete fails here rather than in someone's upgrade.

Proven by mutation — each turns it red:

| Mutation | Failed check |
|---|---|
| parent deleted before its children | `the task types are deleted after the rows that reference them` |
| `virus` left in the manifest's `tables` | `the manifest no longer expects a virus table` |
| retired entry removed | `the manifest declares it retired, with a reason` |
| `clamav` restored in the boot menu | `no clamav kernel argument is emitted` |
| `Schema::dropTable('virus')` removed | `the virus table is dropped in the same step` |
| historical `CREATE TABLE \`virus\`` rewritten | `the historical CREATE TABLE \`virus\` step is still present` |

Local: `sh tests/run-all.sh` → 278 passed, 0 failed. Both `phpstan` passes clean (the `schema.php` baseline counts move 119 → 122 and 375 → 376 for the step's three `self::$DB` calls and one `$this->schema[]`; bumped, not regenerated).

## Ordering

This can land on its own. FOGProject/fos#28 removes `bin/fog.av` and ClamAV from the buildroot config — once this merges nothing sets `mode=clamav`, so that half is dead code and does not need to be atomic with this.

No route change, so nothing for `OpenAPI::document()` and nothing downstream in FogApi.


mastacontrola:
Rebased onto `working-1.6` after #1604 landed ADR 0038's schema steps 402–405. The removal step is renumbered **402 → 406**, `FOG_SCHEMA` is 406, and the constraint-map counts take both changes (121 → 120 relationships, 105 enabled, polymorphic subtotal 7 → 6).

Re-proved against a **fresh** clone of the live 1.6 database rather than assuming the renumber was cosmetic — naive order still refused with errno 1451, step 406 runs clean, taskLog row still survives. Local suite 280 passed / 0 failed, both phpstan passes clean.

mastacontrola:
## Heads-up: `fogproject / upgrade rehearsal` was already red on `working-1.6`

The failure this PR hit is **not** its own. #1604 added four enabled relationships to `commons/schema-constraints.php` for the group-grant tables and did not update `tests/fixtures/upgrade-rehearsal-baseline.txt`. Its own run of that check failed too, and it merged anyway — so the gate has been red on the branch since, and #1601, #1603, #1607 and #1609 have landed past it.

Attributed rather than assumed: a detached worktree at `origin/working-1.6` (`e531c2f2a`), carrying none of this branch, produces the identical diff:

```
-  constraints declared and applicable here: 76
-  constraints actually present:             74
+  constraints declared and applicable here: 80
+  constraints actually present:             78
   MISSING:                                  2
     fk_hostMAC_hmHostID                      CASCADE     orphan rows: 0
     fk_nfsGroupMembers_ngmGroupID            RESTRICT    orphan rows: 1
```

This branch contributes **nothing** to either figure — the relationship it removes, `virus.vHostMAC -> hostMAC.hmMAC`, was `class => 'poly'`, `action => 'none'` with no `enabled` key, so it was never among the constraints the rehearsal applies. The two `MISSING` entries are unchanged, which is the part of the fixture that carries meaning.

Refreshed here (b3351c29f) rather than in its own PR because the gate blocks every branch against `working-1.6`, and landing it inside one that is otherwise green unblocks them a queue cycle sooner.

mastacontrola:
## Update: the baseline refresh landed independently — this PR no longer carries it

`f550a404c` ("Move the rehearsal baseline for the four group-grant foreign keys") moved 76 → 80 / 74 → 78 on `working-1.6` directly. Merging base in absorbed it, so the hunk described in the comment above is **gone from this PR's diff** — `git diff origin/working-1.6...HEAD` no longer touches `tests/fixtures/upgrade-rehearsal-baseline.txt`.

The attribution in that comment still stands as the record of *why* the gate was red; only the "refreshed here" part is superseded. `fogproject / upgrade rehearsal` is now green on `working-1.6` and on every open PR.

### Also in this push (`e7cfd7d19`)

Merged `working-1.6` again after #1616 and #1617. One conflict, in `System.php`: base bumped `FOG_BCACHE_VER` 356 → 357 on the line below my `FOG_SCHEMA` change. Resolved by keeping both — `FOG_SCHEMA` 406, `FOG_BCACHE_VER` 357. Nothing here touches JS or CSS, so the cache constant needed no further bump.

**The step number is unchanged.** Base is still at `FOG_SCHEMA` 405 with `// 405` as its highest label, so 406 remains correct and the live-database proof from the previous rebase still applies.

Re-verified after the merge: `tests/run-all.sh` → **284 passed, 0 failed**; both `phpstan` passes clean against a cleared `/tmp/phpstan`.

### One process note

`fogproject / upgrade rehearsal` is a **required context** as of today. It had been reporting since `602489a20` (2026-08-31) without ever being required — the ruleset was last edited the day before — which is how #1604 merged with it red. Documented in #1619.

PR #1630, opened 2026-09-01
Remember column widths per user, and show description on hover

mastacontrola:
Two grid rules, both applied in `registerTable()` so they hold on every list page rather than being decided page by page.

## Column widths now persist

Dragging a column border, or double-clicking one to fit, survived only until the page was reloaded — the widths lived in `fogColWidthStore`, in memory, per page load. They now ride the table's DataTables state, the same object already carrying column order, visibility, page length and sort, so `stateSaveCallback` writes them to localStorage **and** to the preference store and they follow the user to another machine.

The open question that kept this in memory was what happens when a table's columns change underneath a stored layout. The answer is the one the saved sort already uses: **store against the column's name, not its position.** An index shifts the moment a column is added or removed and every remembered width then lands on the wrong column. A name the table no longer has is ignored; a column the store has never seen takes its own freshly measured width (`fogRestoredColWidths` already did this for the Responsive case); a layout that goes bad ages out with the rest of the state after `stateDuration`.

`makeColumnsResizable()` therefore takes the table API as an optional argument — it is what resolves `data-dt-column` to a column name. Without one it still resizes, it just forgets on reload.

A save is triggered only by the two **gestures** that change a width. Not by a sizing pass: those fire on every ajax load, tab show and window resize, and would turn a layout nobody touched into a stream of preference writes.

## Description is a row tooltip, not a column

A description is prose about the whole record, routinely longer than every other cell put together, so as a column it either dictates the table's widths or is clipped to an ellipsis that says nothing. It is now hidden and set as the row's native `title` from the shared `rowCallback` — native rather than a Bootstrap tooltip because this runs for every row of every draw and a 500-row grid has no business building 500 tooltip instances.

**Hidden rather than dropped, deliberately.** The column still rides the request, so the global search box and the Filter button still match on a description: `FOGManagerController::filter()` resolves each searchable column out of `request.columns`, which carries invisible columns too, so removing it would have quietly taken description search with it. It is kept out of the Column Visibility picker on all three toolbars, because "sometimes a column, sometimes a tooltip" is a worse answer than either one.

Column visibility is itself part of the saved state and the state wins over the column definition, so an existing saved layout is corrected on load. Without that, anyone who had already loaded one of these grids would keep the column and see no change at all.

Plugin Management carried hardcoded percentage widths in which Description was 32%; the five that remain are rebalanced to sum to 100.

## Behaviour given up

Sorting a grid by description, which needed a visible column. Searching on it is unaffected.

## Verification

- `tests/description-tooltip-and-column-widths.test.php` is new and pins both rules. **Every assertion in it was made to fail by mutating the source** before it was kept — ten mutations, ten reds, then restored.
- The three new pure functions were exercised directly under node: a description column is hidden and marked `noVis` while its neighbours are untouched, a state saved while the column was visible comes back hidden with the rest of the layout intact, and a width key follows the column rather than the position.
- Full local suite green (`tests/*.test.php`, `tests/*.test.sh`), both `phpstan` passes green after `rm -rf /tmp/phpstan`.
- `FOG_BCACHE_VER` 358 → 359.

Not verified in a browser — that needs a deploy. Worth checking there: hover a row on Hosts and on Plugins for the description, then drag a column border, reload, and confirm the width came back.

## Downstream

None. No route, schema or class-list change.

### DIFF
--- a/packages/web/management/js/fog/fog.common.js
+++ b/packages/web/management/js/fog/fog.common.js
@@ -462,7 +462,46 @@ function fogHideDescriptionState(indexes, state) {
  *
  * @return {void}
  */
+var fogPrefInFlight = {},
+  fogPrefPending = {};
+
 function fogPrefStore(key, value, cb) {
+  // AT MOST ONE WRITE PER KEY IN FLIGHT, and the newest value wins.
+  //
+  // One user gesture routinely produces SEVERAL saves of the same key: a grid
+  // writes its whole state on every column-sizing pass and every redraw, and
+  // DataTables fires those itself while a column is being resized. Sent
+  // concurrently they are answered in whatever order the server finishes
+  // them, and the row keeps whichever ANSWERS last rather than whichever was
+  // SENT last.
+  //
+  // Measured on the lab, one double-click-to-fit on the host list:
+  //
+  //     send 1012   send 1012   send 1522
+  //       ok 1012     ok 1522     ok 1012
+  //
+  // -- the stale write answered last, so the page showed the fitted layout
+  // while the server kept the old one, and the next page load undid the fit.
+  // It reads exactly like the layout being ignored, which is why it is worth
+  // the queue: the write succeeded, twice, with the wrong value.
+  //
+  // So while a key's write is in flight, later values for it are held rather
+  // than sent, and only the newest is sent when that one completes. Anything
+  // superseded in between is dropped -- it was never going to be the final
+  // answer. Per KEY, not globally: two different preferences have no ordering
+  // relationship and must not wait on each other.
+  if (fogPrefInFlight[key]) {
+    // Callbacks accumulate rather than replace. A caller that passed one is
+    // waiting to hear what happened, and dropping its callback with its value
+    // would leave it waiting forever; they are all told the outcome of the
+    // write that actually lands, which is the one whose value stuck.
+    fogPrefPending[key] = {
+      value: value,
+      cbs: (fogPrefPending[key] ? fogPrefPending[key].cbs : []).concat(cb ? [cb] : [])
+    };
+    return;
+  }
+  fogPrefInFlight[key] = true;
   $.ajax({
     url: fogApiBase() + 'system/userpref/' + encodeURIComponent(key),
     type: 'POST',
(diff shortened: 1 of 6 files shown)


mastacontrola:
## The restore *was* being undone — but the widths weren't the bug

Verified against the 1.6 lab by serving this branch's `fog.common.js` into the live page (nothing written to the webroot). The width restore works; a **double-click-to-fit** then didn't survive a reload. Watching the wire for that one gesture — the number is the fitted column's share ×10000:

```
send 1012   send 1012   send 1522
  ok 1012     ok 1522     ok 1012
```

Three POSTs to one preference key in flight at once: two carrying the layout as it was (DataTables writes the whole state on every column-sizing pass and every redraw, and fires those itself during a resize), then ours carrying the fit. Each is correct when sent. The server keeps whichever **answers** last — a stale one. The page showed the fit, the server held the old value, the next load put the old value back.

Nothing errored. The write succeeded, twice, with the wrong value.

**`fogPrefStore()` now serializes per key** (`73c25ec3f`): while a key's write is in flight, later values are held and only the newest is sent when it completes. Same gesture after the fix:

```
send 502   ok 502   send 1148   ok 1148
```

Three details that are load-bearing and invisible in the code: the key is released in `complete` not `success` (else one 500 wedges that preference for the page's life); queued callbacks accumulate rather than replace (else a caller waits forever); and the value sent is the newest, not the next in line.

**This was never width-specific.** Every grid's column order, visibility, page length and sort goes through the same call, and any two overlapping saves had the same coin flip. Widths only made it visible because one gesture produces three saves carrying two different values.

## Measured after the fix

| | |
|---|---|
| 7 grids swept (image, snapin, group, user, plugin, storagenode, active tasks) | drag → reload → autosize: layout restored and held on all seven |
| double-click-to-fit → reload | exact same shares back |
| real viewport 1500 → 1200 → 900 → 1500 | proportions held, exact layout back at 1500 |
| sidebar collapse | proportions held; a column Responsive brings back takes its own natural width, by design |
| description column / row tooltip | column absent, row carries `title="Created by FOG Reg on…"` |

`tests/userpref-writes-are-serialized.test.sh` executes `fogPrefStore` against a stub `$.ajax` that never answers on its own, so the test controls completion order — a grep cannot tell a correct queue from one that releases on `success` or replaces its callbacks. Five checks, each made to fail by mutation before it was kept.

Full local suite and both phpstan passes green.

Issue #198, opened 2017-10-23
Reliable identification of unique client systems


Sebastian-Roth:
So far FOG used to identify clients via MAC address. Though MACs are pretty much unique all over the world we see more and more people using shared USB NICs to clone systems. This needs addressing.

First we added System UUID as an identifier but those [turned out](https://forums.fogproject.org/topic/10977/error-restoring-images-clients-having-identity-crisis) to be pretty [much unreliable](https://forums.fogproject.org/topic/10976/mac-address-problem-maybe).

Sebastian-Roth:
Working towards a way to identify unique client systems reliably we need to combine the information we have. Though I don't think it's wise to add another field to the DB. Better if we just add the logic to the PHP code to use different identifiers - one after the other in case we still get too many results from the DB.

Sebastian-Roth:
**Nevertheless IMHO this means that we have to give up on MAC address being a unique identifier in the DB** - which will be a major change. Therefore I started a new branch called [unique-identifier](https://github.com/FOGProject/fogproject/tree/unique-identifier) to work on this subject.

astrugatch:
I am running into this exact issue. My work around has been to issue usb NICs to individual machines, but this is just that, a work around. If you need testers for this feature please let me know.

Sebastian-Roth:
We had a long and intensive discussion on this topic and thought that essentially the hard disk is what makes a client a client from the FOG point of view. So I came up with some code to read HDD serial and WWN from the disk firmware. Adding that code to iPXE turned out to be a dead end as there is no disk driver/subsystem in iPXE yet and it would have been a huge task.

I then decided to go back to what we have, UUID, System Serial Number and Mainboard Serial Number. This should be good enough to uniquely identify most machines. Anyhow it's heaps better than what we have right now.

One big step is done. Added most of the code to identify hosts into the FOG PHP code and the FOS scripts. So this will be ready to test fairly soon.

Sebastian-Roth:
People might be able to change their system UUID in some cases:
https://forums.mydigitallife.net/threads/bios-tools.529/page-19
https://kb.stonegroup.co.uk/how-to-change-the-dmi-information-a-stone-asus-based-system_357.html
https://kb.stonegroup.co.uk/how-to-change-the-dmi-information-on-a-stone-intel-nuc-system_412.html

Sebastian-Roth:
[...]

PR #1668, opened 2026-09-02
Identify a booting machine by its firmware, MAC first (#198)

mastacontrola:
Closes nothing yet; this is the first shippable step on #198, and it ships in **log mode**.

## What changes

A booting machine now reports its SMBIOS UUID, system serial, board serial and chassis asset tag (iPXE's `${uuid}`, `${serial}`, `${board-serial}`, `${asset}`, no iPXE change) alongside its MACs. FOG can identify it by them. The MAC stays the identity; the firmware overrides it only when it is sure and only when told to. Full reasoning in ADR 0039.

- **`SmbiosIdentity`** (new, pure): canonicalization, placeholder rules (known firmware defaults plus any single repeated character, which catches the MSI `FFFFFFFF-...` UUID that got the 2018 attempt reverted), and `pick()`: one point per field compared per field, sole top scorer wins, a tie is no answer, the asset tag breaks ties but cannot win alone.
- **`FOG_HOST_IDENTIFY_SMBIOS`** (schema 413, General Settings): `off` / `log` / `enforce`, shipped as **`log`**. The MAC decides and every disagreement is written to the web server error log as `FOG host identity (log): ...`, so a site reads what its vendors' firmware really reports before trusting it. `enforce` lets a unique firmware match win, still logged.
- **Boot fills the inventory row.** A UI/CSV/API-created host gets a row with just the identity fields on its first PXE boot, so it is firmware-findable before it has ever imaged. Existing empty fields are filled; a serial FOS stored is never overwritten, and a disagreement between iPXE and dmidecode is logged. The UUID keeps its validated-replace rule (Aisle 019).
- **Indexes** on the four inventory columns. The lookup runs on every `boot.php` request and the table had none.
- The decision runs on the iPXE boot path, not in `getHostItem()`: FOS's inventory POST reuses the same field names for base64 payloads.

## What does not change

- FOS and the FOG client: the deploy path already hands FOS the stored primary MAC on the kernel line, so resolving once at iPXE boot is enough.
- Anyone on `off`, or on a schema that has not applied 413 yet (missing row reads as `off`).

## Tests

- `tests/smbios-host-identity.test.php`: 33 checks over the five rules. Each rule was proven red by mutation (tie rule, per-field compare, asset-alone) before restore.
- iPXE golden: gains exactly the three param lines in 48 places, nothing else.
- Both phpstan passes green; `tests/run-all.sh` 305 passed.

## Lab

The VirtualBox EFI VM accepts crafted DMI values through `VBoxManage setextradata` on both the `pcbios` and `efi` device paths; byte-for-byte round trip of `${serial}` / `${board-serial}` / `${asset}` against the inventory row is verified after this is deployed to the lab server (results to follow in a comment).

## Downstream

No route change. FogApi unaffected.

https://claude.ai/code/session_01LVm9aM4BHJywMoh1eaJMpZ


mastacontrola:
Lab, first half (no deploy needed): the VirtualBox EFI VM `fos-deploy-test` was given crafted DMI values (`VBoxManage setextradata` on the `efi` device path; all-digit values need the `string:` prefix or the VM refuses to power on) and PXE-booted to the FOG menu. The **existing** `boot.php` UUID write stored `01234567-89ab-cdef-0123-456789abcdef` in `inventory.iSystemUUID`, byte-for-byte as set. So the rig reaches iPXE's `${uuid}`, and iPXE does not byte-swap it on this firmware.

Second half needs this branch on the lab server: boot the same VM and read the `FOG host identity` lines in the web server error log, which print the iPXE-reported `${serial}` / `${board-serial}` / `${asset}` next to the stored dmidecode values (`7XPOC01`, `/7XPOC01/CN1234567890AB/`, `70638948` are what the VM now reports). Then task a FOS inventory to see dmidecode's rendering of the same three.

mastacontrola:
Lab, second half, with #1669 and #1672 merged and the installer re-run: the EFI VM PXE-booted through the new `default.ipxe` and 15 seconds after selecting PXEv4 its inventory row held `sysserial=7XPOC01`, `mbserial=/7XPOC01/CN1234567890AB/`, `caseasset=70638948`, byte-for-byte what was set in DMI. So `${serial}`, `${board-serial}` and `${asset}` map to system serial, baseboard serial and chassis asset tag as documented, arrive unaltered, and the boot-time fill works on a row whose serials were blank. Log mode is live on the lab server.

PR #1669, opened 2026-09-02
default.ipxe sends the firmware identity params (#198)

mastacontrola:
Follow-up to #1668. That PR added the three new params to the script `boot.php` generates, but the **first** `boot.php` request of every PXE boot comes from the installer-written `default.ipxe` (`lib/common/functions.sh`, `configureDefaultiPXEfile`), and that request is the one the host is identified on. It still sent only `sysuuid`.

Found on the lab: with #1668 deployed, the VM reached the FOG menu with its serials still blank, while a direct POST of the same values to the live `boot.php` stored all three byte-for-byte. Selecting a menu item afterward does send them (that path uses the generated script), which is what hid it in the golden.

- `default.ipxe` now carries `param sysserial ${serial}`, `param mbserial ${board-serial}`, `param caseasset ${asset}` beside `sysuuid`, above the NIC chain that short-circuits to `:bootme`.
- `tests/netboot-host.test.sh` pins each param on the written file and its position, the same way it pins the Secure Boot params. Proven red by dropping one (2 of 34 failed), green restored.

Needs an installer run to reach `/tftpboot/default.ipxe`; `copybacktrunk.sh` does not write it.

https://claude.ai/code/session_01LVm9aM4BHJywMoh1eaJMpZ


mastacontrola:
Second commit on this branch: the param block boot.php appends to every **menu item** was built with `'\n'` in single quotes, so iPXE got one line, `param sysuuid ${uuid}\nparam sysserial ...\ngoto bootme`, and stored the whole remainder as the `sysuuid` value. Pre-existing for the UUID (which is why only the first request of a boot ever recorded it); the new params were riding inside the same broken value. Seen live on the lab VM: picking "Client System Information" booted FOS and stored nothing. Now real newlines; the iPXE golden changes in exactly those 46 places.

PR #1673, opened 2026-09-02
Match optional route segments longest-first, so the search box works again

mastacontrola:
**Symptom:** the sidebar universal search returns nothing for any term.

**Cause:** the AltoRouter → FastRoute swap (98159c0a4) changed how an optional segment is matched. AltoRouter ran one regex per route with a greedy optional group; FastRoute expands `[/{limit}]` into two routes and its default parser tries the **shorter** first. The two orders agree until a capture that may contain a slash precedes the optional, and the table has exactly one such route: `/[search|unisearch]/[*:item]/[i:limit]?`. Dispatching `/unisearch/a/5` gave `item="a/5"` and no limit, so the box searched for `%a/5%`.

Probe against the current dispatcher, before this change:

```
/unisearch/a/5 => {"item":"a/5"}          # expected {"item":"a","limit":"5"}
```

**Fix:** `FOG\Router\LongestFirstRouteParser` extends FastRoute's `Std` and reverses the expansion, so the longer form is tried first and falls through to the shorter one when it does not match — the greedy-optional semantics every route in the table was written against. `Route::newDispatcher()` is now the one place the dispatcher is built.

**Test:** `route-table-translates-to-fastroute.test.php` was building its own default dispatcher, which is why it could not see this; it now dispatches through `Route::newDispatcher()`. Added checks: `/unisearch/a/5` → item `a`, limit `5`; `/unisearch/a` → no limit; `/unisearch/a/b` → item `a/b` (non-numeric tail is still the term, as under AltoRouter). Mutation: with `Std` swapped back in, 2 of 35 checks fail. Full suite 306/306, phpstan clean on both configs.

https://claude.ai/code/session_0166dqQEjAs9fhqUw5zCjvxM

### DIFF
--- a/bin/psr4-scan.php
+++ b/bin/psr4-scan.php
@@ -146,6 +146,7 @@ const TABLE = [
     'Route' => 'Router',
     'OpenAPI' => 'Router',
     'HTTPResponseCodes' => 'Router',
+    'LongestFirstRouteParser' => 'Router',
     'Authorization' => 'Auth',
     'CSRF' => 'Auth',
     'SiteScope' => 'Auth',
--- a/packages/web/src/Router/LongestFirstRouteParser.php
+++ b/packages/web/src/Router/LongestFirstRouteParser.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * FastRoute's route parser, with optional segments tried longest-first.
+ *
+ * PHP version 7.4+
+ *
+ * @category LongestFirstRouteParser
+ * @package  FOGProject
+ * @author   Tom Elliott <tommygunsster@gmail.com>
+ * @license  http://opensource.com/licenses/gpl-3.0 GPLv3
+ * @link     https://fogproject.org/
+ */
+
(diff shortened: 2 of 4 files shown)


mastacontrola:
Live reproduction on the 1.6 lab server (1.6.0-beta, schema 413), read-only, before this fix:

```
POST /fog/unisearch/a/5    → 200 application/json  {"_query":"a\/5","_lang":{...}}   # no result buckets
POST /fog/unisearch/fog/5  → 200                    {"_query":"fog\/5", ...}          # same empty shape
POST /fog/unisearch/fog    → 200                    ipxe/setting/user/ldap/oidcgroup buckets populated
```

Browser: the select2 box fires `POST /fog/unisearch/a/5` (with the CSRF header `bootstrap-csrf.js` adds), gets 200, renders "No results found". No console errors, no PHP warnings — the mis-parse is silent. The nginx access log shows the same 233–239 byte empty responses for every term typed from a real browser this morning.

PR #1727, opened 2026-09-08
See an indented `use` in the bare-core-reference gate

mastacontrola:
Follow-up to #1726, where this gate reported a false positive.

## What was wrong

`tests/no-bare-core-references.test.php` collected a file's imports with `/^use\s+([^;]+);$/m` — anchored at the start of a line. That is correct for the unbraced `namespace X;` form every core file uses, and wrong for a file declaring **more than one** namespace: PHP requires braces there, so every import inside one is indented and the regex never saw it. The gate then reported a properly imported class as a bare reference.

Hit on #1726. That branch's new test needs two `filter_input()` overrides in two different namespaces plus global code in one file, which forces bracketed blocks — it is the only such file in the tree — and the gate told it that its correctly imported `RegisterClient` was a bare reference. It was worked around there with FQCN; this fixes the gate.

## Why not just allow leading whitespace

That is one character, and it trades the false positive for a false negative — by a route worth spelling out, because it is not the obvious one.

A trait import inside a class body is *also* an indented `use X;`. `packages/web/src/Base/FOGPage.php:45` really does say `use FOGPageRender;`. The gate does **not** report that line itself — a reference is only counted after `new`/`extends`/`implements`/`instanceof` or before `::`, and a trait import is none of those. The damage is elsewhere: `$known` suppresses **every** occurrence of a short name in the file, so binding the trait import would also silence a `new FOGPageRender()` further down, which the gate *does* detect.

Checked both ways on a file carrying the trait import and the `new` together:

| | reports `new FOGPageRender()` |
|---|---|
| tokeniser (this PR) | yes |
| whitespace-tolerant regex | **no** — binds `FOGPageRender`, reports nothing |

Only brace depth separates a trait import from a namespace import, so import collection moves to the tokeniser the reference scan already uses. Closure `use (...)` is skipped because it binds variables, not names; an alias binds the alias.

Group imports (`use A\{B, C};`) and `use function`/`use const` are deliberately not handled — `git grep` finds none in the tree, and a group import would bind nothing and so read as a bare reference, which fails in the safe direction. Both are noted in the docblock for whoever lands the first one.

## Self-check

Nothing in the tree exercises this. Every tracked file today puts its imports at column 0, so a regression that stopped seeing indented ones would scan all 389 files and report a clean pass — the same failure mode the daemon-count assertion above it already guards against. So the fix carries a self-check over a synthetic token stream covering the indented import, the alias, the trait import and the closure.

Verified by mutation, not by assertion: binding trait imports, and ignoring indented imports, each fail it by name.

## Not fixed here

`$ns` (the `namespace` capture) is assigned and never read — pre-existing dead code, and the docblock's claim that "a name reached through a namespace the file's own `namespace` resolves" is not reported describes behavior that does not exist. Left alone rather than widened into this diff; happy to take it in a separate pass.

## Checks

No behavior change for any file currently in the tree — 389 scanned, 0 bare references, unchanged. Suite 340/340, both `phpstan` passes clean.

https://claude.ai/code/session_01Q97JnkpnQTUoBZuwZzckjn

### DIFF
--- a/tests/no-bare-core-references.test.php
+++ b/tests/no-bare-core-references.test.php
@@ -114,6 +114,175 @@ if (count($daemons) < 10) {
 }
 
 $skipTokens = [T_WHITESPACE, T_COMMENT, T_DOC_COMMENT];
+
+/**
+ * The short names a file already resolves through a namespace-level import.
+ *
+ * Tokenised, for the same reason the reference scan below is: a regex has to
+ * anchor somewhere, and the one this replaced anchored `use` at the start of
+ * a line. That is correct for the unbraced `namespace X;` form every core
+ * file uses, and wrong for a file declaring MORE THAN ONE namespace -- PHP
+ * requires braces there, so every import inside one is indented, and the gate
+ * reported properly imported classes as bare references (GH-1726).
+ *
+ * Simply allowing leading whitespace would have traded that false positive
+ * for a false NEGATIVE, by a route worth spelling out because it is not the
+ * obvious one. A trait import inside a class body is also an indented
+ * `use X;` -- packages/web/src/Base/FOGPage.php:45 really does say
+ * `use FOGPageRender;` bare. The gate does NOT report that line itself: a
+ * reference is only counted after new/extends/implements/instanceof or
+ * before `::`, and a trait import is none of those. The damage is that
+ * $known suppresses EVERY occurrence of a short name in the file, so
+ * binding the trait import would also silence a `new FOGPageRender()`
+ * further down -- a reference the gate does detect. Verified both ways
+ * against a file carrying the trait import and the `new` together: the
+ * tokeniser still reports the `new`; a whitespace-tolerant regex binds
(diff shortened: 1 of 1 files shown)


mastacontrola:
Correction to the "Not fixed here" section of the description above: it says `$ns` was left alone. That is wrong about this PR's own diff — the commit *does* remove it, as a side effect of replacing the regex block it lived in. Apologies for the misleading note; I wrote it without re-reading the diff.

What that section should have said is that removing the capture left the exemption it was reaching for still unimplemented — the header has promised "a name reached through a namespace the file's own `namespace` resolves" is not reported for as long as this file has existed, and it never was. That is a latent false positive of exactly the kind this PR fixes: scanned files really do declare namespaces (`FOG\Base`, `FOG`, `FOGTest`), so a same-namespace bare reference resolves at runtime but would be reported.

Following up in a separate PR that implements it and puts both halves of the gate behind self-checks.

